PR #29, opened 2020-03-20
add closing backquote in docstring

ho-oto:


Jutho:
Thanks for catching (and correcting) this.

PR #34, opened 2020-06-23
correct typo

rveltz:


Jutho:
Thanks!

rveltz:
I had been seeing that one way too much :D

Jutho:
:-) I don't use the verbosity very often, as my linear problems are inside steps of an optimization problem, and otherwise I have too much output. 

I don't know if spellchecking can be included as part of the tests :-).

Should I tag a new version soon? It's no effort to make a simple patch release, so I will probably do that. I don't have any changes in the pipeline in the near future (unless there turn out to be bugs).

PR #39, opened 2020-10-13
Allow eigsolve to return less than `howmany` values

mhauru:
As discussed.


Jutho:
Somehow my mails failed to arrive in my email application for the whole afternoon, so I only now see this. Anyway, look's good and so I merge.

PR #47, opened 2021-04-19
Fix typos in documentation

kyungminlee:


Jutho:
Thanks for this; this is an amazing work. You must really have studied all these files in detail. As you can see, I am not a native English speaker, and US versus British English (and something non-existent in between) is all the same to me.

PR #48, opened 2021-04-19
Fix `rayleighquotient` for `LanczosFactorization`

kyungminlee:
The last element of β should be ignored in the symmetric tridiagonal Rayleigh quotient

### DIFF
--- a/src/krylov/lanczos.jl
+++ b/src/krylov/lanczos.jl
@@ -20,7 +20,7 @@ Base.eltype(::Type{<:LanczosFactorization{<:Any,S}}) where {S} = S
 
 basis(F::LanczosFactorization) = length(F.V) == F.k ? F.V :
     error("Not keeping vectors during Lanczos factorization")
-rayleighquotient(F::LanczosFactorization) = SymTridiagonal(F.αs, F.βs)
+rayleighquotient(F::LanczosFactorization) = SymTridiagonal(F.αs, F.βs[1:end-1])
 residual(F::LanczosFactorization) = F.r
 @inbounds normres(F::LanczosFactorization) = F.βs[F.k]
 rayleighextension(F::LanczosFactorization) = SimpleBasisVector(F.k, F.k)


src/krylov/lanczos.jl
```diff
@@ -20,7 +20,7 @@ Base.eltype(::Type{<:LanczosFactorization{<:Any,S}}) where {S} = S
 
 basis(F::LanczosFactorization) = length(F.V) == F.k ? F.V :
     error("Not keeping vectors during Lanczos factorization")
-rayleighquotient(F::LanczosFactorization) = SymTridiagonal(F.αs, F.βs)
```
Jutho, line 23: This was done for efficiency reasons. It's impossible to create `SymTridiagonal(F.αs, view(F.βs, 1:end-1))` because `SymTridiagonal` expects both vectors to be of the same type. However, `SymTridiagonal(F.αs, F.βs[1:end-1])` creates a copy on every call, which is not necessary and therefore suboptimal. It seems that the `SymTridiagonal` is happy accepting a second vector which is one element too long, and then just ignores it (perhaps precisely for this purpose). So I would leave this as is.

kyungminlee: Makes sense.
The reason I created this PR is that, the resulting `SymTridiagonal` could not be passed to `eigen`. The `LinearAlgebra` module in Julia 1.6 requires `length(ev) == length(dv) - 1`.

https://github.com/JuliaLang/julia/blob/7cbb02d47539d162b57b41398d13d5da2ad42de2/stdlib/LinearAlgebra/src/lapack.jl#L3822
But it seems that the current master branch allows that extra offdiagonal term at the end, so we could just wait for the next release.
 https://github.com/JuliaLang/julia/blob/b67d706f6553cf572fa1aa547ba8586c4eee4b04/stdlib/LinearAlgebra/src/lapack.jl#L3827

Jutho: Indeed, I remember this, that was partly the reason for having my own routines. I was certainly planning to report this at some point as a Julia issue, but probably forgot. Glad that it will be fixed.

PR #53, opened 2022-03-18
Reuse initial apply in expintegrator

emstoudenmire:
At the beginning of `expintegrator`, apply is called to apply the linear map for the purpose of determining the type `T` of other variables. However, the result is not used later in the code. This PR uses the result to start building the vectors `w`, saving one call to apply while giving exactly the same results. This could lead to a significant performance increase in cases where only a few Krylov vectors are computed, such as for small time steps.

### DIFF
--- a/src/matrixfun/expintegrator.jl
+++ b/src/matrixfun/expintegrator.jl
@@ -136,7 +136,7 @@ function expintegrator(A, t::Number, u::Tuple, alg::Union{Lanczos,Arnoldi})
     # initial vectors
     w = Vector{typeof(w₀)}(undef, p + 1)
     w[1] = w₀
-    w[2] = one(T) * Au₀
+    w[2] = one(T) * Au₀ # reuse the result of apply computed earlier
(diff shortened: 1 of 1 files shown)


Jutho (changes requested):
Thanks for detecting this. I'll be happy to merge, but may I suggest the following replacement instead:
```julia
    w[1] = w₀
    w[2] = mul!(similar(w₀), Au₀, one(T))
    for j in 2:p
```

emstoudenmire:
I'm happy to make that change. One question though: in changing the loop range to start at 2, the code at the bottom of that loop won't get run for the j=1 case. Is that a concern?

Jutho:
You are of course absolute right; I was clearly not very well awake when I wrote that. I will just merge the current version. Thanks again.

emstoudenmire:
Great! I just went ahead and did change that one line to be `w[2] = mul!(similar(w₀), Au₀, one(T))` as you mentioned.

PR #56, opened 2022-04-28
Eigsolve AD

lkdvos:
Almost working version of the eigsolve rrules.
Seems to be failing only for real matrices, where comparison with rrule for eigen does not give equal results.
Specifically, it seems that the current rrule has the exact same real part, but different imaginary part.

### DIFF
--- a/src/eigsolve/eigsolve.jl
+++ b/src/eigsolve/eigsolve.jl
@@ -304,4 +304,4 @@ function conjwhich(which::Symbol)
     end
 end
 
-conjwhich(s::EigSorter) = EigSorter(conj(s.by), s.rev)
+conjwhich(s::EigSorter) = EigSorter(x -> s.by(conj(x)), s.rev)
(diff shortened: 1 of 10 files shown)


Jutho:
Cool; I'll try to review as soon as possible.

maartenvd:
> Seems to be failing only for real matrices, where comparison with rrule for eigen does not give equal results.

is the problem the sorting, or is there something mathematically wrong?

lkdvos:
> > Seems to be failing only for real matrices, where comparison with rrule for eigen does not give equal results.
> 
> is the problem the sorting, or is there something mathematically wrong?

I am just guessing, but I don't see what else could be different for real matrices

src/eigsolve/eigsolve.jl
```diff
@@ -292,3 +293,15 @@ function eigsort(which::Symbol)
...
+    if which == :LI
+        return :SI
+    elseif which == :SI
+        return :LI
+    else
+        return which
+    end
+end
+
+conjwhich(s::EigSorter) = EigSorter(conj(s.by), s.rev)
```
Jutho, line 307: `s.by` can be a function, so I don't see how that would work? Do you mean something like `x->s.by(conj(x))` ?

lkdvos: ah, my mistake, I only  tested with ```:LM``` so this escaped from my attention, should be fixed now

Jutho:
Closed in favor of #85

PR #65, opened 2022-12-01
VectorInterface

lkdvos:
rewrite algorithms in terms of VectorInterface methods.
* Some additional cleanup can be done in terms of scalartype
* algorithms can be written in terms of bangbang-methods, possibly allowing tuples as vectors?


Jutho:
Hi Lukas; thanks for this great work. I was indeed considering switching to bangbang methods when adopting VectorInterface, so if you have time to try this out and see if there would be any unexpected consequences or complications, feel free to do so ...

lkdvos:
@Jutho I seem to be stuck with the geneigsolve for out-of-place vectors, although I can't seem to locate what's going wrong. It produces matrices that are not positive definite, such that geneigh! fails, but I have absolutely no clue why this would be different for the out-of-place methods.
I might look into it later this week, but any help is definitely welcome.
(tests are ran by just commenting out modules "svdsolve.jl" and "expintegrator.jl")

src/adrules/linsolve.jl
```diff
@@ -58,26 +58,28 @@ function ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode},
 
     # f defines a linear map => pullback defines action of the adjoint
     (y, f_pullback) = rrule_via_ad(config, f, x)
-    fᴴ(xᴴ) = f_pullback(xᴴ)[2]
+    fᴴ(xᴴ) = add(zerovector(x), f_pullback(xᴴ)[2])
```
lkdvos, line 61: I'm still not sure about this way of implementing it. This is the equivalent of trying to reconstruct the vector type, but I feel like this should really be the responsability of the user to make sure that the rrule of the function outputs something sensible. Alternatively, it would be nice to have the option of specifying this manually

lkdvos:
Some small remaining remarks:
- I would like to still restructure some of the tests, such that there is less duplicated code in the tests, and less tests run multiple times.
- Somehow this still had the eigenvalue ad rules that are not in the main branch. It might be better to add these in a separate PR, which immediately adds the ADrules as a package extension.
- There are no tests for the eigsolve ad, but this should probably also happen in the separate pr as mentioned above

lkdvos:
So, if the lights turn green, this should be ready for review.

I excluded AD rules from the new functionality, because we should probably rewrite that section anyways, using Package extensions.

Otherwise, I think everything should now work with `VectorInterface`, and the tests have been reorganised to extensively test everything with regular vectors, and then do some small additional tests with `MinimalVec`-type vectors.

In principle I think we can also remove/deprecate `RecursiveVec` and `InnerProductVec`, but this would probably require some deprecation update first?

src/adrules/linsolve.jl
lkdvos: I didn't realise that this was already changed. I think it's better to keep the old behaviour for now. The reason for the new implementation was that the tangent types of ChainRules are not necessarily compatible with VectorInterface, and this bypasses that by forcing the tangent type to be the same as the input vector type. However, it seems needlessly lossy to do the addition every time, allocating 2 extra vectors, so for now let's just assume that the user writes their code such that the AD rule of applying the operator yields a compatible vector type.

Jutho (commented):
For the tests, I have mixed feelings. The reduced runtime is very good, but it seems this comes with quite a bit of near-trivial code duplication.

Jutho:
Nightly failures seem due to Zygote failing?

lkdvos:
I think so, it also seems to vary quite a bit depending on what version of nightly, so I think we can ignore this. We'll refactor that part anyways when we move the AD support to a package extension I guess.

Jutho (approved):

PR #83, opened 2024-04-02
AD extension package

lkdvos:
This moves the AD rules into a package extension.
The implementation for `linsolve` comes from the current master
The implementation for `eigsolve` comes from one of the open PRs (#56 )

## TODO
- [ ] look into setting up tests in more uniform way
- [ ] use ChainRulesTestUtils
- [ ] add `svdsolve` AD rules
- [ ] add `expsolve` AD rules


lkdvos:
I think in any case we should look for a solution to determine how the `rrule` is computed, even just changing the tolerances should be accessible. 

Also, I think @pbrehmer has an implementation for the `svdsolve`, so I'll tag him here :)

PR #84, opened 2024-04-23
Add svdsolve AD rule

pbrehmer:
This PR adds a reverse-rule for `svdsolve` which implements the truncated SVD adjoint recently derived in [this pre-print](https://arxiv.org/abs/2311.11894). See also PR #83.


pbrehmer:
For the first commit I mostly copied the code from quantumghent/PEPSKit.jl#15 and adapted it a bit to KrylovKit. As of now, this is untested but the PEPSKit version runs correctly and produces results that are consistent with the TensorKit SVD reverse-rule. However, this PR does need a bit of help to get it fully running @lkdvos.

lkdvos:
Great, thanks!
( I had in mind merging it into #83 first, such that I can immediately make it part of the package extension :smile: )
I'll try and have a look at this later this week!

Jutho:
Am I correct that this only works for the case where `A` is an explicit matrix?

pbrehmer:
> Am I correct that this only works for the case where `A` is an explicit matrix?

Yes, I forgot to add that. I wasn't sure how KrylovKit internally handles the abstract vector types and what would be the best way to rewrite things like `U = hcat(lvecs...)` as well as the matrix multiplications that are performed.

pbrehmer:
I think this should do the job? I hope this is mergeable now.

Jutho:
Ok, I pushed some updates to the ad-extension branch, but now there is a conflict again. Can I resolve it (this will also push to your master) or do you want to rebase from your side?

pbrehmer:
> Can I resolve it (this will also push to your master) or do you want to rebase from your side?

Yes sure, you can go ahead and resolve it (and push).

PR #94, opened 2024-10-07
Small fixes and doc updates

lkdvos:


Jutho:
From Lapack's `dhseqr.f` docstring:
```fortran
*     INFO  (output) INTEGER
*             =  0:  successful exit
*           .LT. 0:  if INFO = -i, the i-th argument had an illegal
*                    value
*           .GT. 0:  if INFO = i, [DHSEQR](https://www.netlib.org/lapack/explore-3.1.1-html/dhseqr.f.html#DHSEQR.1) failed to compute all of
*                the eigenvalues.  Elements 1:ilo-1 and i+1:n of WR
*                and WI contain those eigenvalues which have been
*                successfully computed.  (Failures are rare.)
```
So we hit a rare failure :smile:. I'll merge.

PR #97, opened 2024-11-05
deprecate RecursiveVec, improve x0 initialisation for GPU

Jutho:


src/eigsolve/eigsolve.jl
```diff
@@ -177,12 +177,14 @@ EigSorter(f::F; rev=false) where {F} = EigSorter{F}(f, rev)
 
 const Selector = Union{Symbol,EigSorter}
 
+# TODO: initialize with `Random.rand!(similar(A, size(A, 1)))` for GPU arrays
```
lkdvos, line 180:
```suggestion
```

lkdvos (approved):

src/eigsolve/eigsolve.jl
Jutho: Why not simply add:
```julia
# TODO: remove TODOs for things that have been done
```

test/nestedtuple.jl
lkdvos, line 1: I think if we deprecate `RecursiveVec`, it might not be a bad idea to keep the name here for now, and only change the tests after the deprecation is removed.

Jutho: I might be wrong, but I don't think that is common practice in e.g. Julia's main repository?

lkdvos: I don't know, I think I once googled "how to deprecate in Julia" and found this (possibly outdated) blogpost:
https://invenia.github.io/blog/2022/06/17/deprecating-in-julia/

Jutho: Ok, that seems to pretty clearly confirm what you say. I guess most of my expertise comes from the Julia < 1 versions, where I think it used to be the case that new deprecations were introduced in going from 0.x to 0.x+1, and then removed in 0.x+2.

PR #99, opened 2024-11-08
fix arnoldi ad rule for degenerate eigsolve

Jutho:


Jutho:
Ok, if CI finishes I think this is ready. It now seems to robustly deal with degenerate cases.

PR #103, opened 2024-11-12
Fix and test for issue 100

### LINKED ISSUE #100: `AbstractArray` vector subtypes get wrongly handled
It seems like whenever the custom vector-type is `<:AbstractArray`, and has `IndexLinear` indexing style, we eagerly assume that it can be treated as if it is an `Array`, taking [this path](https://github.com/Jutho/KrylovKit.jl/blob/1b175b7bd57a9535e26b738dbc7403ec30aee444/src/orthonormal.jl#L297) in the code for `basistransform!`. I think it would probably be more reasonable to have `<:DenseArray` or even just `<:Array` there, as we are no longer adhering to the `VectorInterface` set of allowed operations there.

For example, `similar` is being used instead of `zerovector` (https://github.com/Jutho/KrylovKit.jl/blob/1b175b7bd57a9535e26b738dbc7403ec30aee444/src/orthonormal.jl#L336), and then we index into these vectors (https://github.com/Jutho/KrylovKit.jl/blob/1b175b7bd57a9535e26b738dbc7403ec30aee444/src/orthonormal.jl#L342), which fails for non-`isbits` types.

Additionally, there really is no guarantee that `AbstractArray`s are efficiently indexable (or the indexing even means the same thing).
For example in MPSKit (where I ran into this issue), we have that `QuasiParticle` states subtype `AbstractVector`, because you can index into them by site, while we also have that they are elements of a vectorspace, for which we overload `VectorInterface`. While this design might be questionable, I would still expect KrylovKit to use the `VectorInterface` methods, instead of the `AbstractVector` methods.

[Edit] Actually, the reason it was going wrong is that in the `Multiline` tests, we used a `RecursiveVec` for the different lines, which I now changed for a regular `Vector`, thus changing this to `<:DenseArray` or `<:Array` also does not suffice.

lkdvos:
This is a more conservative approach to using the multithreaded kernels, which could even be reenabled by users by overloading the `_use_multithreaded_array_kernel`.

Fixes #100

### DIFF
--- a/src/orthonormal.jl
+++ b/src/orthonormal.jl
@@ -62,6 +62,19 @@ LinearAlgebra.mul!(y, b::OrthonormalBasis, x::AbstractVector) = unproject!!(y, b
 
 const BLOCKSIZE = 4096
 
+# helper function to determine if a multithreaded approach should be used
+# this uses functionality beyond VectorInterface, but can be faster
+_use_multithreaded_array_kernel(y) = false
+function _use_multithreaded_array_kernel(::Array{T}) where {T<:Number}
+    return isbitstype(T) && get_num_threads() > 1
+end
+function _use_multithreaded_array_kernel(::Type{<:Array{T}}) where {T<:Number}
+    return isbitstype(T) && get_num_threads() > 1
+end
+function _use_multithreaded_array_kernel(::OrthonormalBasis{T}) where {T}
+    return _use_multithreaded_array_kernel(T)
+end
+
 """
     project!!(y::AbstractVector, b::OrthonormalBasis, x,
         [α::Number = 1, β::Number = 0, r = Base.OneTo(length(b))])
@@ -127,8 +140,7 @@ function unproject!!(y,
                      α::Number=true,
                      β::Number=false,
                      r=Base.OneTo(length(b)))
-    if y isa AbstractArray && !(y isa AbstractGPUArray) && IndexStyle(y) isa IndexLinear &&
-       get_num_threads() > 1
+    if _use_multithreaded_array_kernel(y)
         return unproject_linear_multithreaded!(y, b, x, α, β, r)
     end
     # general case: using only vector operations, i.e. axpy! (similar to BLAS level 1)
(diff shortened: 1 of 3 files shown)


src/orthonormal.jl
```diff
@@ -62,6 +62,19 @@ LinearAlgebra.mul!(y, b::OrthonormalBasis, x::AbstractVector) = unproject!!(y, b
 
 const BLOCKSIZE = 4096
 
+# helper function to determine if a multithreaded approach should be used
+# this uses functionality beyond VectorInterface, but can be faster
+_use_multithreaded_array_kernel(y) = false
```
Jutho, line 67: Maybe, in order to avoid duplication of cases:
```julia
_use_multithreaded_array_kernel(y) = _use_multithreaded_array_kernel(typeof(y))
_use_multithreaded_array_kernel(::Type) = false
function _use_multithreaded_array_kernel(::Type{<:Array{T}}) where {T<:Number}
     return isbitstype(T) && get_num_threads() > 1
 end
 function _use_multithreaded_array_kernel(::Type{<:OrthonormalBasis{T}}) where {T}
     return _use_multithreaded_array_kernel(T)
 end
```

Jutho:
Looks good. Could you also check if this is the only reason why GPUArraysCore.jl was introduced as dependency, and if so, remove it from Project.toml?

Project.toml
```diff
@@ -1,10 +1,9 @@
 name = "KrylovKit"
 uuid = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
 authors = ["Jutho Haegeman"]
-version = "0.8.2"
+version = "0.8.3"
```
Jutho, line 4: With the new functionality of the other PR, I would probably go to 0.9. Unless you need this as quickly as possible (I guess the other one would take a few more changes and one more CI run) and want to already tag this right now. I can then also change to 0.9 after merging #105 .

Jutho:
Feel free to merge when you see fit.

Project.toml
lkdvos: I think it would be convenient for me if it was 0.8.3, since the RecursiveVec thing broke things... In the end, it does not matter too much, I changed this here to test it out with MPSKit

PR #105, opened 2024-11-13
Implement support for real linear maps

Jutho:
This PR refactors the existing `realeigsolve`, and similarly implements a `reallinsolve`, to deal with so called `real linear maps`, i.e. maps that are additive f(x+y) = f(x)+f(y) and homogeneous f(a * x) = a * f(x), but only with respect to real scalars a. It can happen that x and y are nonetheless complex vectors, but f(x) is only real linear because it for example depends on `conj(x)`. Indeed, any real linear map can be represented as `f(x) = A * x + B * conj(x)` where `A` and `B` are complex matrices, but we want to work with an implicit representation of course.

The only change necessary to consider `x` and `y` as elements of a real vector space (with a real dimension that is twice as large as the original complex dimension) is to replace their complex inner product with its real part. Then, we can use the standard algorithms. For `realeigsolve`, we of course can only have real eigenvalues with corresponding eigenvectors, as complex linear combinations are not allowed (and would in fact yield wrong results).

TODO: tests

### DIFF
--- a/docs/src/man/matfun.md
+++ b/docs/src/man/matfun.md
@@ -1,5 +1,5 @@
-# Functions of matrices and linear maps
-Applying a function of a matrix or linear map to a given vector can in some cases also be
+# Functions of matrices and linear operator
+Applying a function of a matrix or linear operator to a given vector can in some cases also be
 computed using Krylov methods. One example is the inverse function, which exactly
 corresponds to what `linsolve` computes: ``A^{-1} * b``. There are other functions ``f``
 for which ``f(A) * b`` can be computed using Krylov techniques, i.e. where ``f(A) * b`` can
--- a/docs/src/man/matfun.md
+++ b/docs/src/man/matfun.md
@@ -1,4 +1,4 @@
-# Functions of matrices and linear operator
+# Functions of matrices and linear operators
 Applying a function of a matrix or linear operator to a given vector can in some cases also be
 computed using Krylov methods. One example is the inverse function, which exactly
 corresponds to what `linsolve` computes: ``A^{-1} * b``. There are other functions ``f``
--- a/docs/src/man/reallinear.md
+++ b/docs/src/man/reallinear.md
@@ -1,0 +1,46 @@
+# Real linear maps
+
+A map $$f: V \to V$$ from some vector space $$V$$ to itself is said to be real linear map if
+it satisfies $$f(\alpha x + \beta y) = \alpha f(x) + \beta f(y)$$ for all $$x, y \in V$$ and
+all $$\alpha, \beta \in \mathbb{R}$$. When $$V$$ is itself a real vector space, this is just
+the natural concept of a linear map. However, this definition can be used even if $$x$$ and
+$$y$$ are naturally represented using complex numbers and arithmetic and also admit complex linear
+combinations, i.e. if $$V$$ is a complex vector space.
+
+Such real linear maps arise whenever `f(x)` involves calling `conj(x)`, and are for example
+obtained in the context of Jacobians (pullbacks) of complex valued functions that are not
+holomorphic.
+
+To deal with real linear maps, one should reinterpret $$V$$ as a real vector space, by
+restricting the possible linear combinations to those with real scalar coefficients, and by
+using the real part of the inner product. When the vectors are explictly represented as
+some `AbstractVector{Complex{T}}`, this could be obtained by explicitly splitting
+them in their real and imaginary parts and stacking those into `AbstractVector{T}` objects
+with twice the original length.
+
+However, KrylovKit.jl admits a different approach, where the original representation of
+vectors is kept, and the inner product is simply replaced by its real part. KrylovKit.jl
+offers specific methods for solving linear systems and eigenvalue systems in this way. For
+linear problems, this is implemented using `reallinsolve`:
+
+```@docs
+reallinsolve
+```
+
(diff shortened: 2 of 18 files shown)


docs/src/man/matfun.md
```diff
@@ -1,5 +1,5 @@
-# Functions of matrices and linear maps
-Applying a function of a matrix or linear map to a given vector can in some cases also be
+# Functions of matrices and linear operator
```
lkdvos, line 1:
```suggestion
# Functions of matrices and linear operators
```

docs/src/man/reallinear.md
```diff
@@ -0,0 +1,46 @@
+# Real linear maps
+
+A map $$f: V \to V$$ from some vector space $$V$$ to itself is said to be real linear map if
```
lkdvos, line 3:
```suggestion
A map $$f: V \to V$$ from some vector space $$V$$ to itself is said to be a real linear map if
```

src/eigsolve/arnoldi.jl
```diff
@@ -268,33 +286,30 @@ The return value is always of the form `vals, vecs, info = eigsolve(...)` with
...
-            end
-        end
-        allreal || throw(ArgumentError("not all first `howmany` eigenvalues are real"))
-    end
-    if converged > howmany
-        while howmany < converged && T[howmany + 1, howmany] == 0
-            howmany += 1
+    realinner = real ∘ inner
+    v₀ = InnerProductVec(x₀, realinner)
+    realA = v -> InnerProductVec(apply(A, v[]), realinner)
```
lkdvos, line 291: How would you feel about avoiding the anonymous functions:
```julia
const realinner = real ∘ inner
const RealVec{V} = InnerProductVec{typeof(realinner),V}
RealVec(v) = InnerProductVec(v, realinner)

apply(A, x::RealVec) = RealVec(apply(A, x[]))
...
```
The main point being that the KrylovKit methods then automatically use the correct functions

(I don't care too much about the names, mostly about the interface)

src/factorizations/arnoldi.jl
```diff
@@ -142,7 +142,7 @@ function initialize(iter::ArnoldiIterator; verbosity::Int=0)
     α = inner(x₀, Ax₀) / (β₀ * β₀)
     T = typeof(α)
     # this line determines the vector type that we will henceforth use
-    v = add!!(zerovector(Ax₀, T), x₀, 1 / β₀)
+    v = add!!(scale(Ax₀, zero(α)), x₀, 1 / β₀)
```
lkdvos, line 145: Is this because `inner` can now return a number which is not the `scalartype` of the vectors? It might be convenient to mention that here, because next time I look at this I will definitely not remember.

Perhaps we should also at some point centralize that logic somewhere, as this is repeated in many places in the code and is a bit messy anyways

test/runtests.jl
```diff
@@ -25,8 +25,8 @@ const mgsr = ModifiedGramSchmidtIR(η₀)
 # Tests
 # -----
 t = time()
-include("factorize.jl")
-include("gklfactorize.jl")
+# include("factorize.jl")
+# include("gklfactorize.jl")
```
lkdvos, line 29: This is presumably accidental?
```suggestion
include("factorize.jl")
include("gklfactorize.jl")
```

lkdvos (commented):
Looks good overall.

I think I would like it if the real vector logic/applying matrices is refactored into its own functions, but otherwise I only found some small typos

src/eigsolve/arnoldi.jl
Jutho: I am definitely in favour of the first three lines. The redefinition of `apply` might be a bit more problematic, as maybe some user is constructing the same type of `RealVec` objects, as this definition of `real ∘ inner` does not create a new function that is intrinsic and private to KrylovKit.jl. That user might then define their map to unpack the vector, and this would lead to a conflict. 

The solution might be to have an internal `_realinner(v,w) = real(inner(v,w))` such that `RealVec{V} = InnerProductVec{typeof(_realinner, V}` is impossible to ever be constructed by an external user, without hijacking our private `_realinner`.

src/factorizations/arnoldi.jl
Jutho: Yes, `scalartype` used to mean basically the scalar field over which the vector space was defined, and the whole real business makes this ambiguous. So indeed, if `T` is simply determined from the inner product of vectors, it will be real for `RealVec`, but the scalartype is actually still complex. Given that `scalartype` is now part of the required interface (even though I don't think we use it), we could also do something like `T = promote(typeof(α), scalartype(Ax₀))`. Although then we have to be careful furtheron, where `T` is also used to initialise the matrix `H` and the other data related to the Krylov subspace, which should be fully determined by the return type of the inner product alone. 

Actually, it seems the GMRES implementation was already doing something similar in its initialisation stage, so that one did not have an issue.

test/runtests.jl
Jutho: Yes. That's me skipping over to the relevant tests 😄

src/eigsolve/arnoldi.jl
lkdvos: that's definitely also fine by me

src/factorizations/arnoldi.jl
Jutho: I've added some comments, so I think this is ok now?

PR #109, opened 2024-12-06
bring LSMR implementation of jutho/krylovkit.jl#46 up to date

VictorVanthilt:
other than change @maartenvd 's code to use the new `apply_normal` and `apply_adjoint` syntax, this PR is equivalent to #46.


VictorVanthilt:
This should now be ready to merge

src/algorithms.jl
```diff
@@ -281,6 +281,54 @@ function GMRES(;
     return GMRES(orth, maxiter, krylovdim, tol, verbosity)
 end
 
+"""
+LSMR(; orth = KrylovDefaults.orth,atol = KrylovDefaults.tol,btol = KrylovDefaults.tol,conlim = 1/KrylovDefaults.tol,
+    maxiter = KrylovDefaults.maxiter,krylovdim = KrylovDefaults.krylovdim,λ = 0.0,verbosity = 0)
```
lkdvos, line 286: I think this docstring could use some formatting improvements and a bit of a clean up 😉

src/algorithms.jl
```diff
@@ -281,6 +281,54 @@ function GMRES(;
...
+
+"""
+struct LSMR{O<:Orthogonalizer,S<:Real} <: KrylovAlgorithm
+    orth::O
+    atol::S
+    btol::S
+    conlim::S
+    maxiter::Int
+    verbosity::Int
+    λ::S
```
lkdvos, line 318: Having unicode as names for fields can turn out to be not entirely convenient, since this means there is no alternative to access that field on a machine/terminal that has no support for this. While I also like to use unicode, I would advise to only do that for internal variable names, or in cases where an alternative alias can be provided.

src/linsolve/linsolve.jl
lkdvos, line 1: Probably it would be good to have an additional mention to the docstring that LSMR requires a different interface for the function handles, since it needs both the regular as well as the adjoint action.

src/linsolve/lsmr.jl
```diff
@@ -0,0 +1,214 @@
...
+            # Compute norms for convergence testing.
+            normAr = abs(ζbar)
+            normx = norm(x)
+
+            # Now use these norms to estimate certain other quantities,
+            # some of which will be small near a solution.
+            test1 = normr / normb
+            test2 = normAr / (normA * normr)
+            test3 = inv(condA)
+
```
lkdvos, line 148: I'm a bit confused by the placement of these variables. Is there a reason to define these before the logger, and to then use `test1`, `test2` and `test3` below with some unexplained conditions? I think it is a bit more readable to keep that together:
`test3 + 1 <= 1` vs `inv(condA) + 1 <= 1` seems like it reads better

src/linsolve/lsmr.jl
```diff
@@ -0,0 +1,214 @@
...
+            # atol = eps,  btol = eps,  conlim = 1/eps.
+
+            if alg.verbosity > 2
+                msg = "LSMR linsolve in iter $topit; step $(length(fact)-1): "
+                msg *= "normres = "
+                msg *= @sprintf("%.12e", normr)
+                @info msg
+            end
+
+            if 1 + test3 <= 1
```
lkdvos, line 164: Can this be re-expressed as `test3 <= eps(one(test3))` ? This is somewhat strange code, because you could easily think that this just means `test3 <= 0`, which would be a different check.

test/linsolve.jl
lkdvos, line 1: I guess you should probably also test a non-square matrix here, since that is the primary use-case of the method?

src/linsolve/lsmr.jl
Jutho: Are these different convergence criteria currently used? If not, I would probably streamline them to be compatible with what we use in the other methods, even if this "reduces" the functionality or options to specify the convergence. I will take a stab at this myself.

VictorVanthilt:
This PR was closed in favor of jutho/KrylovKit.jl/#111

PR #111, opened 2025-01-04
LSMR further updates

Jutho:
@VictorVanthilt , this is based on your PR and contains an updated LSMR implementation. While it is quite different from the original implementation, your PR was still very helpful to finally push me to study and implement this. Major changes as compared to PR #109 :
* Because LSMR cannot live up to its promises (minimise ||x|| over all solutions if A has a nontrivial nullspace) when a nonzero starting guess is provided, the ability to specify a starting guess has been disabled. The doc string still mentions how you could manually take a starting guess into account. Comments about this are welcome
* Removed the GKLIterator dependence, which requires to restart after `krylovdim` steps, which cannot become very large. Furthermore, also restarting has the issue of the starting vector. The new implementation is almost exactly as discussed in the LSMR paper.
* Because of the several interface differences with other `linsolve` calls (no starting vector, a linear map for which also the adjoint needs to be defined, possibility to "regularise" the solution with a parameter lambda) I decided to move this to a separate (and thus new) function, called `lssolve` (for solving least squares problems).


Jutho:
There is one remaining todo (which could also be dealt with at a later time) and one interface questions:
* the residual r = b - A * x is not by default determined in the LSMR implementation, only its norm is. I currently explicitly compute the residual at the very end (which requires one more matrix-vector product), but probably the residual can be build during the iterative process in a short recurrence style in the same way as x itself is constructed.
*  there are a number of interesting measures. norm(r) is interesting but cannot be expected to become very small for an overconstrained problem. The actual measure for the convergence is norm(A'*r) when lambda = 0, and norm(A'*r - lambda^2 x) otherwise. An estimate for this value is currently stored in `info.normres`. So in this case, `info.normres` is not `norm(info.residual)`. Ideas and comments to improve this situation are welcome.

Issue #114, opened 2025-01-18
Issue with `//` in `expintegrator` in KrylovKit.jl v0.9


mtfishman:
This call to `//` is causing issues in our TDVP code: https://github.com/Jutho/KrylovKit.jl/blob/v0.9.0/src/matrixfun/expintegrator.jl#L207 (which was introduced in https://github.com/Jutho/KrylovKit.jl/pull/106, and included in KrylovKit.jl v0.9).

It looks like `q` is a floating point value, so `//` fails. This reproduces the issue:
```julia
julia> using KrylovKit

julia> exponentiate(randn(10, 10), 1.2, randn(10); krylovdim=2)
ERROR: MethodError: no method matching //(::Int64, ::Float64)
The function `//` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  //(::Number, ::Complex)
   @ Base rational.jl:101
  //(::LinearAlgebra.UniformScaling, ::Number)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.11.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/uniformscaling.jl:273
  //(::AbstractArray, ::Number)
   @ Base rational.jl:104
  ...

Stacktrace:
 [1] expintegrator(A::Matrix{Float64}, t::Float64, u::Tuple{Vector{Float64}, Vector{Float64}}, alg::Arnoldi{ModifiedGramSchmidt2, Float64})
   @ KrylovKit ~/.julia/packages/KrylovKit/gFbJF/src/matrixfun/expintegrator.jl:207
 [2] expintegrator
   @ ~/.julia/dev/KrylovKit/src/matrixfun/expintegrator.jl:102 [inlined]
 [3] expintegrator(::Matrix{Float64}, ::Float64, ::Vector{Float64}; kwargs::@Kwargs{krylovdim::Int64})
   @ KrylovKit ~/.julia/packages/KrylovKit/gFbJF/src/matrixfun/expintegrator.jl:98
 [4] expintegrator
   @ ~/.julia/packages/KrylovKit/gFbJF/src/matrixfun/expintegrator.jl:94 [inlined]
 [5] #exponentiate#100
   @ ~/.julia/packages/KrylovKit/gFbJF/src/matrixfun/exponentiate.jl:83 [inlined]
 [6] top-level scope
   @ REPL[23]:1

julia> pkgversion(KrylovKit)
v"0.9.0"
```

[...]

PR #115, opened 2025-01-20
Various fixes for v9

Jutho:
Fixes issue #114 and some other issues had so far not been detected.


Jutho:
I think this is good to go, not sure if you still want to review @lkdvos . Let's merge and tag this asap, as the current `expintegrator` in KrylovKit v0.9 is broken and I guess more issue reports will be coming in.

PR #117, opened 2025-02-05
Remove unused code (reference to undefined `msg`)

lkdvos:
This should in principle be caught by tests...

### DIFF
--- a/src/eigsolve/arnoldi.jl
+++ b/src/eigsolve/arnoldi.jl
@@ -348,12 +348,6 @@ function _schursolve(A, x₀, howmany::Int, which::Selector, alg::Arnoldi)
     β = normres(fact)
     tol::eltype(β) = alg.tol
 
-    if eltype(fact) <: Real && krylovdim < 4
-        error("krylov dimension should be at least 4 to avoid getting stuck in the Arnoldi process in real arithmetic")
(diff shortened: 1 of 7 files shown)


src/eigsolve/arnoldi.jl
```diff
@@ -348,6 +348,12 @@ function _schursolve(A, x₀, howmany::Int, which::Selector, alg::Arnoldi)
     β = normres(fact)
     tol::eltype(β) = alg.tol
 
+    if eltype(fact) <: Real && krylovdim < 4
+        error("krylov dimension should be at least 4 to avoid getting stuck in the Arnoldi process in real arithmetic")
+    elseif krylovdim < 2
+        error("krylov dimension should be at least 2 to avoid getting stuck in the Arnoldi process in complex arithmetic")
+    end
+
```
lkdvos, line 356: Does this mean that we cannot factorize small matrices using KrylovKit? It would have some additional complications for working with TensorKit, where the blocks could end up quite small.

Jutho:
Ok, I think this is ready @lkdvos . Will you merge?
Do not yet tag the new version, I will also add a `reallssolve` implementation.

lkdvos:
I don't have the rights to approve this, but good to go for me!

Issue #110, opened 2024-12-10
`svdsolve` rrule error when not all values are converged


lkdvos:
In one of the PEPSKit testruns, we are running into an issue where the within the rrule for `svdsolve`, the eigensolver is failing to converge all requested values. In this case, the output `vals` and `vecs` are not of the expected length, leading to an out of bounds error.

The stacktrace is found [here](https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/12263340824/job/34214671852?pr=97)
The offending line is [this line](https://github.com/Jutho/KrylovKit.jl/blob/4d2a06f898fe1f87c8de240ed7c82c3591b9c934/ext/KrylovKitChainRulesCoreExt/svdsolve.jl#L237) .
There is already a warning being thrown, but I guess we have to limit `n` to the size of the computed vals and vecs.

lkdvos:
Yeah, that's exactly what it's hitting. The log is a bit cryptic (blame Zygote), but this is relevant:
```julia
[ Info: CTMRG conv 4:	obj = +9.192864957180e-01	err = 4.1934293618e-11	time = 0.75 sec
┌ Warning: Invariant subspace of dimension 7 (up to requested tolerance `tol = 1.0e-8`), which is smaller than the number of requested eigenvalues (i.e. `howmany == 14`); setting `howmany = 7`.
└ @ KrylovKit ~/.julia/packages/KrylovKit/xccMN/src/eigsolve/arnoldi.jl:350
┌ Warning: `svdsolve` cotangent linear problem (7) returns unexpected result
└ @ KrylovKitChainRulesCoreExt ~/.julia/packages/KrylovKit/xccMN/ext/KrylovKitChainRulesCoreExt/svdsolve.jl:240
┌ Warning: `svdsolve` cotangent linear problem (14) returns unexpected result
└ @ KrylovKitChainRulesCoreExt ~/.julia/packages/KrylovKit/xccMN/ext/KrylovKitChainRulesCoreExt/svdsolve.jl:240
```

Jutho:
Ok that is very weird. The different eigenvectors of the linear operator are supposed to be of the form

$$((\lambda_i - A)^{-1}\Delta_i , e_i)$$

with $e_i$ the length $n$ unit vector in the $i$ direction, and this for $i=1,...,n$. The initial vector is given by $(0, e_1 + e_2 + ... + e_n)$. That should have support in each of the eigenspaces.

[...]

PR #123, opened 2025-03-11
try more stable svd ad rule computation

Jutho:
This is supposed to improve both the error/warning messaging and the actual computation of the AD data for svd. If this helps in practice, I might want to do the same for the eigenvalue ad rule. This is ready to be tried, it should pass the tests.

This will hopefully fix/improve #110 : @leburgel


leburgel:
I tried it out this PR on QuantumKitHub/PEPSKit.jl#151, where the test failures are being caused by #110. This PR does not solve the original problem of #110, but combined with QuantumKitHub/PEPSKit.jl#155 where I scaled the rrule alg tolerance according to the smallest magnitude singular value, this seems to resolve all remaining issues, i.e. I don't get any warnings about the cotangent linear problem returning an unexpected result and everything converges smoothly.

It's hard to assess the improvement in the actual computation of the svd data, but just from the fact that the spurious warnings are gone this seems perfect from my end.

PR #124, opened 2025-03-19
Add Two-sided Arnoldi method

tipfom:
I've added an implementation of the Two-sided Krylov-Schur Restarted Arnoldi method in Ref. [1, Alg. 2] to KrylovKit.jl. The implementation is not done yet and needs some improvements (e.g., unit tests and an update to the convergence criterion). 

It is currently only tested for finding the smallest real eigenvalue :SR for dense Julia matrices. The interface is as close as possible to the other methods, i.e., for a dense matrix A calling 
`eigsolve(A, collect(adjoint(A)), v0, w0, 4, :SR, BiArnoldi(; krylovdim=20, verbosity=10, maxiter=100))`
yields the 4 smallest real value left and right eigenvalues, vectors and convergence information.

Are you interested in merging this into KrylovKit.jl? If so, what should I additionally add (next to tests and the improved convergence criterion)?

I'm interested in using this method downstream (in ITensors.jl) and hence incorporating this into the package would be very much appreciated. I'm willing to do the extra work required, if it is not too much.

Thanks for your work!

[1] Krylov-Schur-Type Restarts for the Two-Sided Arnoldi Method, Ian N. Zwaan and Michiel E. Hochstenbach

### DIFF
--- a/src/eigsolve/biarnoldi.jl
+++ b/src/eigsolve/biarnoldi.jl
@@ -123,8 +123,8 @@ function _schursolve(A, AH, v₀, w₀, howmany::Int, which::Selector, alg::BiAr
             # Step 2 and 3 - Correct H, K and the residuals using the oblique projection
 
             # Compute the projections W* residual(V) and V* residual(W)
-            Wv = zeros(krylovdim)
-            Vw = zeros(krylovdim)
+            Wv = zeros(eltype(rV), krylovdim)
+            Vw = zeros(eltype(rV), krylovdim)
             for i in eachindex(Wv)
                 Wv[i] = dot(fact.W[i], rV)
                 Vw[i] = dot(fact.V[i], rW)
@@ -200,8 +200,8 @@ function _schursolve(A, AH, v₀, w₀, howmany::Int, which::Selector, alg::BiAr
             Vv = -adjoint(Q[:, 1:keep]) * MWv
             Ww = -adjoint(Z[:, 1:keep]) * MVw
 
-            _H[1:keep, 1:keep] += Vv * _h[1:keep]'
-            _K[1:keep, 1:keep] += Ww * _k[1:keep]'
+            _H[1:keep, 1:keep] += Vv * transpose(_h[1:keep])
+            _K[1:keep, 1:keep] += Ww * transpose(_k[1:keep])
 
             # newresidual = (I - Vm Vm*) oldresidual = (I - Vl Q1 Vm*) oldresidual = oldresidual + Vl Q1 Q_1^* MWv = oldresidual + Vl Q1 Vv
             Q1Vv = Q[:, 1:keep] * Vv
--- a/src/eigsolve/biarnoldi.jl
+++ b/src/eigsolve/biarnoldi.jl
@@ -1,0 +1,253 @@
+function eigsolve(A, AH, v₀, w₀, howmany::Int, which::Selector, alg::BiArnoldi;
+                  alg_rrule=alg)
+    S, Q, T, Z, fact, converged, numiter, numops = _schursolve(A, AH, v₀, w₀, howmany,
+                                                               which, alg)
+
+    howmany′ = howmany
+    if eltype(T) <: Real && howmany < length(fact) && T[howmany + 1, howmany] != 0
+        howmany′ += 1
+    elseif size(T, 1) < howmany
+        howmany′ = size(T, 1)
+    end
+    if converged > howmany
+        howmany′ = converged
+    end
+
+    TT = view(T, 1:howmany′, 1:howmany′)
+    SS = view(S, 1:howmany′, 1:howmany′)
+    valuesT = schur2eigvals(TT)
+    valuesS = schur2eigvals(SS)
+
+    # Compute eigenvectors
+    VS = view(Q, :, 1:howmany′) * schur2eigvecs(SS)
+    VT = view(Z, :, 1:howmany′) * schur2eigvecs(TT)
+    vectorsS = let B = basis(fact)[1]
+        [B * v for v in cols(VS)]
+    end
+    vectorsT = let B = basis(fact)[2]
+        [B * v for v in cols(VT)]
+    end
+
+    residualsS = let r = residual(fact)[1]
+        [scale(r, last(v)) for v in cols(VS)]
+    end
+    residualsT = let r = residual(fact)[2]
+        [scale(r, last(v)) for v in cols(VT)]
+    end
+
+    normresidualsS = [normres(fact)[1] * abs(last(v)) for v in cols(VS)]
+    normresidualsT = [normres(fact)[2] * abs(last(v)) for v in cols(VT)]
+
+    if (converged < howmany) && alg.verbosity >= WARN_LEVEL
+        @warn """Arnoldi eigsolve stopped without convergence after $numiter iterations:
+        * $converged eigenvalues converged
+        * norm of residuals = $(normres2string(normresidualsS))
+        * number of operations = $numops"""
+    elseif alg.verbosity >= STARTSTOP_LEVEL
+        @info """Arnoldi eigsolve finished after $numiter iterations:
+        * $converged eigenvalues converged
+        * norm of residuals = $(normres2string(normresidualsS))
+        * number of operations = $numops"""
+    end
+    return (valuesS, valuesT),
+           (vectorsS, vectorsT),
+           (ConvergenceInfo(converged, residualsS, normresidualsS, numiter, numops), ConvergenceInfo(converged, residualsT, normresidualsT, numiter, numops))
+end
+
+function _schursolve(A, AH, v₀, w₀, howmany::Int, which::Selector, alg::BiArnoldi)
+    krylovdim = alg.krylovdim
+    maxiter = alg.maxiter
+    howmany > krylovdim &&
+        error("krylov dimension $(krylovdim) too small to compute $howmany eigenvalues")
+
+    ## FIRST ITERATION: setting up
+    numiter = 1
+    # initialize arnoldi factorization
+    iter = BiArnoldiIterator(A, AH, v₀, w₀, alg.orth)
+    fact = initialize(iter; verbosity=alg.verbosity)
+    numops = 1
+    sizehint!(fact, krylovdim)
+    βv, βw = normres(fact)
+    tol::eltype(βv) = alg.tol
+
+    # allocate storage
+    HH = fill(zero(eltype(fact)), krylovdim + 1, krylovdim)
+    KK = fill(zero(eltype(fact)), krylovdim + 1, krylovdim)
+    QQ = fill(zero(eltype(fact)), krylovdim, krylovdim)
+    ZZ = fill(zero(eltype(fact)), krylovdim, krylovdim)
+
+    # initialize storage
+    K = length(fact) # == 1
+    converged = 0
+    local S, T, Q, Z
+    while true
+        βv, βw = normres(fact)
+        K = length(fact)
+
+        if (βv <= tol || βw <= tol) && K < howmany
+            if alg.verbosity >= WARN_LEVEL
+                msg = "Invariant subspace of dimension $K (up to requested tolerance `tol = $tol`), "
+                msg *= "which is smaller than the number of requested eigenvalues (i.e. `howmany == $howmany`)."
+                @warn msg
+            end
+        end
+        if K == krylovdim || (βv <= tol && βw <= tol) || (alg.eager && K >= howmany) # process
+
+            # Step 1
+            _H = view(HH, 1:K, 1:K)
+            _K = view(KK, 1:K, 1:K)
+            Q = view(QQ, 1:K, 1:K)
+            Z = view(ZZ, 1:K, 1:K)
+            _h = view(HH, K + 1, 1:K)
+            _k = view(KK, K + 1, 1:K)
+
+            copyto!(Q, I)
+            copyto!(Z, I)
+
+            copyto!(_H, rayleighquotient(fact)[1])
+            copyto!(_K, rayleighquotient(fact)[2])
+
+            rV, rW = residual(fact)
+            normalize!(rV)
+            normalize!(rW)
+
+            # Compute the oblique Projection
+            # TODO: This should reuse the old projection after restarting
+            M = fill(zero(eltype(fact)), krylovdim, krylovdim)
+            for i in axes(M, 1)
+                for j in axes(M, 2)
+                    M[i, j] = dot(fact.W[i], fact.V[j])
+                end
+            end
+
+            # Step 2 and 3 - Correct H, K and the residuals using the oblique projection
+
+            # Compute the projections W* residual(V) and V* residual(W)
+            Wv = zeros(krylovdim)
+            Vw = zeros(krylovdim)
+            for i in eachindex(Wv)
+                Wv[i] = dot(fact.W[i], rV)
+                Vw[i] = dot(fact.V[i], rW)
+            end
+
+            MWv = inv(M) * Wv
+            MVw = inv(M') * Vw
+
+            _H[:, end] += MWv * βv
+            _K[:, end] += MVw * βw
+
+            for i in eachindex(Wv)
+                rV -= fact.V[i] * MWv[i]
+                rW -= fact.W[i] * MVw[i]
+            end
+
+            # Step 5 - Compute dense schur factorization
+            S, Q, valuesH = hschur!(_H, Q)
+            T, Z, valuesK = hschur!(_K, Z)
+
+            # Step 6 - Order the Schur decompositions
+            by, rev = eigsort(which)
+            pH = sortperm(valuesH; by=by, rev=rev)
+            pK = sortperm(valuesK; by=by, rev=rev)
+
+            S, Q = permuteschur!(S, Q, pH)
+            T, Z = permuteschur!(T, Z, pK)
+
+            # Partially Step 7 & 8 - Correction of hm and km
+            _h = mul!(_h, view(Q, K, :), βv)
+            _k = mul!(_k, view(Z, K, :), βw)
+
+            converged = 0
+            # TODO: this needs refinement
+            while converged < length(fact) && abs(_h[converged + 1]) <= tol &&
+                      abs(_k[converged + 1]) <= tol
+                converged += 1
+            end
+            if eltype(T) <: Real &&
+               0 < converged < length(fact) &&
+               T[converged + 1, converged] != 0
+                converged -= 1
+            end
+
+            if converged >= howmany || (βv <= tol && βw <= tol)
+                break
+            elseif alg.verbosity >= EACHITERATION_LEVEL
+                @info "Arnoldi schursolve in iteration $numiter, step = $K: $converged values converged, normres = $(normres2string(abs.(_h[1:howmany])))"
+            end
+        end
+
+        if K < krylovdim # expand
+            fact = expand!(iter, fact; verbosity=alg.verbosity)
+            numops += 1
+        else # shrink
+            numiter == maxiter && break
+
+            # Determine how many to keep
+            keep = div(3 * krylovdim + 2 * converged, 5) # strictly smaller than krylovdim since converged < howmany <= krylovdim, at least equal to converged
+            while keep < krylovdim &&
+                (isapprox(valuesH[pH[keep]], conj(valuesH[pH[keep + 1]]); rtol=1e-3) ||
+                 isapprox(valuesK[pK[keep]], conj(valuesK[pK[keep + 1]]); rtol=1e-3))
+                # increase the number of eigenvalues kept such that the eigenspace is not fractured at a eigenvalue pair
+                keep += 1
+            end
+
+            # Setp 10 & 11 - Correct the kept part of H and K and the residual 
+
+            # We know that 
+            #   Vm* residual(V) = Q_1* (Vl* residual(V)) = Q_1* Vl* (residual - Vl Ml^-1 Wl* residual) = -Q_1* Ml^-1 Wl* residual = -Q_1* MWv 
+            # as Vl*Vl = Id and Vl* residual = 0
+
+            Vv = -adjoint(Q[:, 1:keep]) * MWv
+            Ww = -adjoint(Z[:, 1:keep]) * MVw
+
+            _H[1:keep, 1:keep] += Vv * _h[1:keep]'
+            _K[1:keep, 1:keep] += Ww * _k[1:keep]'
+
+            # newresidual = (I - Vm Vm*) oldresidual = (I - Vl Q1 Vm*) oldresidual = oldresidual + Vl Q1 Q_1^* MWv = oldresidual + Vl Q1 Vv
+            Q1Vv = Q[:, 1:keep] * Vv
+            Z1Ww = Z[:, 1:keep] * Ww
+
+            for i in eachindex(Q1Vv)
+                rV -= fact.V[i] * Q1Vv[i]
+                rW -= fact.W[i] * Z1Ww[i]
+            end
+
+            βpv = norm(rV)
+            βpw = norm(rW)
+
+            _h .*= βpv
+            _k .*= βpw
+
+            # Restore Arnoldi form in the first keep columns; this is not part of the original paper
+            _restorearnoldiformandupdatebasis!(keep, _H, Q, _h, rayleighquotient(fact)[1],
+                                               fact.V, rV, βpv)
+            _restorearnoldiformandupdatebasis!(keep, _K, Z, _k, rayleighquotient(fact)[2],
+                                               fact.W, rW, βpw)
+
+            # Shrink Arnoldi factorization
+            fact = shrink!(fact, keep; verbosity=alg.verbosity)
+            numiter += 1
+        end
+    end
+
+    return S, Q, T, Z, fact, converged, numiter, numops
+end
+
+function _restorearnoldiformandupdatebasis!(keep, H, U, f, rq, B, r, βr)
+    @inbounds for j in 1:keep
+        H[keep + 1, j] = f[j]
+    end
+    @inbounds for j in keep:-1:1
+        h, ν = householder(H, j + 1, 1:j, j)
+        H[j + 1, j] = ν
+        H[j + 1, 1:(j - 1)] .= 0
+        lmul!(h, H)
+        rmul!(view(H, 1:j, :), h')
+        rmul!(U, h')
+    end
+    copyto!(rq, H) # copy back into fact
+
+    # Update B by applying U
+    basistransform!(B, view(U, :, 1:keep))
+    return B[keep + 1] = scale!!(r, 1 / βr)
+end
--- a/src/factorizations/biarnoldi.jl
+++ b/src/factorizations/biarnoldi.jl
@@ -1,0 +1,187 @@
+# biarnoldi.jl
+mutable struct BiArnoldiFactorization{T,S} <: KrylovFactorization{T,S}
+    k::Int # current Krylov dimension
+    V::OrthonormalBasis{T} # basis of length k
+    W::OrthonormalBasis{T} # basis of length k
+    H::Vector{S} # stores the Hessenberg matrix in packed form
+    K::Vector{S} # stores the Hessenberg matrix in packed form
+    rV::T # residual
+    rW::T # residual
+end
+
+Base.length(F::BiArnoldiFactorization) = F.k
+Base.sizehint!(F::BiArnoldiFactorization, n) = begin
+    sizehint!(F.V, n)
+    sizehint!(F.W, n)
+    sizehint!(F.H, (n * n + 3 * n) >> 1)
+    sizehint!(F.K, (n * n + 3 * n) >> 1)
+    return F
+end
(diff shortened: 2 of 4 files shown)


Jutho:
Thanks for this; it seems like an impressive amount of work. I will have to look deeper into this, since I am on my way out of the office right now. Feel free to remind me if you haven't heard back in a few days.

A first comment: look at `svdsolve` and `lssolve` for the interface I use for methods that require both the linear map and its adjoint. 

Related to this: since this algorithm can not have the exact same arguments and outputs as the other `eigsolve` routines, I would pack it under a different name, such as `bieigsolve` or `eigsolve2`. At some point, we were actually discussing having a new routine like this, when we believed that to implement the pullback of `eigsolve`, we would actually also need the corresponding left eigenvectors. However, since I then found a way to compute the pullback without needing the left eigenvectors, we halted the discussion of implementing such a method, however, I am still very supportive of having it.

Jutho:
`dot` should just be `inner`. `normalize!` could be `scale!(v, 1/norm(v))`; the problem is probably that `normalize!` from LinearAlgebra.jl calls `rmul!` or something, which is indeed not defined for `MinimalVec`.

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,286 @@
...
+            _k = view(KK, K + 1, 1:K)
+
+            copyto!(Q, I)
+            copyto!(Z, I)
+
+            copyto!(_H, rayleighquotient(fact)[1])
+            copyto!(_K, rayleighquotient(fact)[2])
+
+            rV, rW = residual(fact)
+            scale!!(rV, 1/βv)
```
Jutho, line 123:
```suggestion
            rv = scale!!(rV, 1/βv)
```

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,286 @@
...
+            scale!!(rW, 1/βw)
```
Jutho, line 124:
```suggestion
            rV = scale!!(rV, 1/βv)
            rW = scale!!(rW, 1/βw)
```

Jutho, line 124:
```suggestion
            rV = scale!!(rV, 1/βv)
            rW = scale!!(rW, 1/βw)
```

src/factorizations/biarnoldi.jl
```diff
@@ -0,0 +1,205 @@
...
+Base.IteratorSize(::Type{<:BiArnoldiIterator}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{<:BiArnoldiIterator}) = Base.EltypeUnknown()
+
+function Base.iterate(iter::BiArnoldiIterator)
+    state = initialize(iter)
+    return state, state
+end
+function Base.iterate(iter::BiArnoldiIterator, state)
+    nr = normres(state)
+    if nr < eps(typeof(nr))
```
Jutho, line 48: Looking at the normres implementation on line 28, I think `nr` here is a tuple with the two norms in the final elements of H and K. Do you want both to be smaller than eps? Or just one of them? I would expect that this immediately throws an error upon the first run.

Jutho:
@tipfom , are you ok with me pushing some changes directly to your branch?

tipfom:
@Jutho Yes, I'll take care of it and also add docstrings :)

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,286 @@
...
+            end
+
+            # Step 5 - Compute dense schur factorization
+            S, Q, valuesH = hschur!(_H, Q)
+            T, Z, valuesK = hschur!(_K, Z)
+
+            # Step 6 - Order the Schur decompositions
+            by, rev = eigsort(which)
+            pH = sortperm(valuesH; by=by, rev=rev)
+            pK = eltype(fact) <: Complex ? sortperm(conj.(valuesK); by=by, rev=rev) : sortperm(valuesK; by=by, rev=rev)
```
Jutho, line 154: I am a bit confused by this line: even if the factorization is real (i.e. computed in real arithmetic), the values can be complex. The only case where you can be sure that all eigenvalues are real, is in the hermitian case. So isn't `sortperm(conj.(valuesK); by=by, rev=rev)` just always what you need?

Jutho:
Ok, I've simplified the BiArnoldi factorization and iteration to simply reuse everything from Arnoldi, as I think that was wat it amounted to. It's easier to maintain the code and make improvements in the future if there is less code duplication.

In `bieigsolve`, I've currently only made the changes required to make it work in combination with the changed factorization type. I will go through that function in more detail but this will have to wait until tomorrow or Wednesday. However, the code is very nicely structured and written, so I don't expect anything stalling this much longer. Hats off for a very nice PR!

src/eigsolve/biarnoldi.jl
tipfom: Yes, you are correct; this line is a very inproper way to achieve what I was trying to do. If the operator, say the matrix A, is purely real, $A = \overline A$, then it has both the eigenvalues $\lambda$ and $\lambda^*$. If only `sortperm(conj.(...),...)` is used, then the algorithm in the purely real case pairs $\lambda$ for $H$ with $\lambda^*$ for $K$, at least in my testings. That's why I added this line. This behaviour should however not be infered from the data type. 

I'll think about it, but there is probably a nice way to solve this issue, which I'm not aware of right now. You can see this issue by supplying the function a random real or random complex matrix and then comparing the left/right eigenvalues produced.

tipfom: This is fixed now

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+    end
+
+    SS = view(S, 1:howmany′, 1:howmany′)
+    TT = view(T, 1:howmany′, 1:howmany′)
+    valuesS = schur2eigvals(SS)
+    valuesT = schur2eigvals(TT)
+
+    if length(valuesS) != length(valuesT)
+        @error "BiArnoldi bieigsolve converged with unequal number of eigenvalues for the left- and right eigenspaces"
+    end
```
lkdvos, line 24: Can this ever happen? These seem to be both determined from the same `1:howmany′, 1:howmany′` view, so should have the same length I think

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+    usedvaluesT = zeros(Bool, length(valuesT))
+    firstunusedT = 1
+    for i in eachindex(matchperm)
+        # as both arrays are sorted roughly similar, tracking the first valid index
+        # changes the scaling of the loop from O(n^2) to rougly O(n)
+        while usedvaluesT[firstunusedT]
+            firstunusedT += 1
+        end
+        for j in firstunusedT:length(valuesT)
+            if !usedvaluesT[j] && isapprox(valuesS[i], conj(valuesT[j]))
```
lkdvos, line 48: Do we have an idea about what the tolerance should be for this `isapprox`?

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+                overlapji = inner(vectorsT[j], vectorsS[i])
+                if !isapprox(norm(overlapji), 0.0)
```
lkdvos, line 50:
```suggestion
                if !isapprox(abs(overlapji), 0.0)
```

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+                    # distribute the weight to both vectors
+                    vectorsS[i] = scale!!(vectorsS[i], 1 / sqrt(overlapji))
+                    vectorsT[j] = scale!!(vectorsT[j], 1 / conj(sqrt(overlapji)))
+                    break
+                end
+            end
+        end
+
+        if matchperm[i] == 0 
+            matchperm = matchperm[1:i-1]
```
lkdvos, line 64: `resize!(matchperm, i-1)`?

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+        * $converged eigenvalues converged
+        * norm of residuals = $(normres2string(normresidualsS))
+        * number of operations = $numops"""
+    elseif alg.verbosity >= STARTSTOP_LEVEL
+        @info """Arnoldi eigsolve finished after $numiter iterations:
+        * $converged eigenvalues converged
+        * norm of residuals = $(normres2string(normresidualsS))
+        * number of operations = $numops"""
+    end
+    return valuesS[1:length(matchperm)], vectorsS[1:length(matchperm)], vectorsT[matchperm],
```
lkdvos, line 91: Here I also think you can get away with a `resize!` for the `valuesS` and `vectorS`.

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+            mul!(_temp, (Z[:, 1:keep])', M)
+            mul!(_M, _temp, Q[:, 1:keep])
+
+            # Shrink Arnoldi factorization
+            fact = shrink!(fact, keep; verbosity=alg.verbosity)
+            numiter += 1
+        end
+    end
+
+    return S, Q, T, Z, fact, converged, numiter, numops
```
lkdvos, line 304: How would you feel about grouping the arguments:
```suggestion
    return (S, Q), (T, Z), fact, converged, numiter, numops
```

This way, `_schursolve` has the same number of output arguments for the different algorithms. This doesn't matter too much, but it can be nice to be able to always assume that `converged` is the 4th output, for example.

lkdvos (commented):
I quickly browsed through this, definitely a very nice PR, thank you for this work!

I left some minor remarks, mostly just cosmetic and minor details about allocations, which I don't think should matter too much. Feel free to ignore if you disagree with anything.

src/eigsolve/biarnoldi.jl
tipfom: You're right! :)

src/eigsolve/biarnoldi.jl
tipfom: I just went with the default, but I'm open to suggestions! I think if they are reasonably converged then sqrt(eps) is sufficient. Maybe one could also implement a match-as-good-as-possible logic, but the isapprox approach worked fine for me. I'll do more tests how this behaves for larger tolerances.

src/eigsolve/biarnoldi.jl
tipfom: Done

src/eigsolve/biarnoldi.jl
tipfom: Done

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,324 @@
...
+    VS = view(Q, :, 1:howmany′) * schur2eigvecs(SS)
+    VT = view(Z, :, 1:howmany′) * schur2eigvecs(TT)
+    vectorsS = let B = basis(fact)[1]
+        [B * v for v in cols(VS)]
+    end
+    vectorsT = let B = basis(fact)[2]
+        [B * v for v in cols(VT)]
+    end
+
+    # we need to match the eigenvalues; sometimes λ and λ* get mismatched,
```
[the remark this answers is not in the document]
tipfom: Yeah, seems reasonable 👍

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,329 @@
...
+            while converged < length(fact)
+                # The authors suggest the convergence should also include the 
+                # 1. a biorthogonality component, i.e., kappa_j / |rho_j| in the paper 
+                #    with kappa_j = norm(w_j* v_j) and rho_j = abs(w_j* A v_j) / kappa_j 
+                # 2. a contribution of the norms of tilde v and tilde w
+
+                # For the first case (1.), we use the Ritz values instead of the Rayleigh quotients 
+                # as suggested by the authors 
+
+                # This is Eq. 10 in the paper
```
Jutho, line 215: I am a bit confused by the following as the paper in Eq.10 defines this quantity for the Ritz vectors, which are the approximate eigenvectors obtained within the subspace. Here, however, you only have the approximate Schur vectors; eigenvectors are computed later. Maybe that is good enough; I do the same in the regular `Arnoldi`.

However, another question is that the expression for `xh` and `xk` are devided by `M[converged+1,converged+1]`, but that is the overlap of the original `v` and `w` vectors before doing the Schur decomposition; M was not transformed with Q and Z.

Jutho:
My apologies for the slow progress from my side; it's a busy week. I've left another question in the comments.

src/eigsolve/biarnoldi.jl
```diff
@@ -0,0 +1,329 @@
...
+
+            numops += 1
+        else # shrink
+            numiter == maxiter && break
+
+            # Determine how many to keep
+            keep = div(3 * krylovdim + 2 * converged, 5) # strictly smaller than krylovdim since converged < howmany <= krylovdim, at least equal to converged
+            while keep < krylovdim &&
+                (isapprox(valuesH[pH[keep]], conj(valuesH[pH[keep + 1]]); rtol=1e-3) ||
+                 isapprox(valuesK[pK[keep]], conj(valuesK[pK[keep + 1]]); rtol=1e-3))
```
Jutho, line 258: Is there a specific reason not to simply check `!iszero(H[keep+1,keep])` as in arnoldi.jl? Also, I think this is only a problem if `H` and `K` are real, i.e. when working in real arithmetic?

tipfom: This would work as well. However, I have the following question regarding this approach: in which sense is it reasonable to only run this check if `eltype(H) <: Real`?  Can't I trick this mechanism by supplying a real matrix `A` as `ComplexF64.(A)` thus changing the `eltype` but not the fundamental properties of the matrix?

Regardless, you may replace this code block by the following or something similar if you like this better:
```julia 
while _H[keep + 1, keep] != 0 || _K[keep + 1, keep] != 0
    # we are in the middle of a 2x2 block; this cannot happen if keep == converged, so we can decrease keep
    # however, we have to make sure that we do not end up with keep = 0
    if keep > 1
        keep -= 1 # conservative choice
    else
        keep += 1
        if krylovdim == 2
            alg.verbosity >= WARN_LEVEL &&
                @warn "Arnoldi iteration got stuck in a 2x2 block, consider increasing the Krylov dimension"
            break
        end
    end
end
```

src/eigsolve/biarnoldi.jl
tipfom: I used the rightmost expressions to determine $\Vert r_j \Vert$ and $\Vert s_j \Vert$, e.g., $\Vert r_j \Vert = \ldots = \Vert \tilde v_{l+1} \Vert \Vert h_l^* c_j \Vert$. Here, both the norm of the projected residual $\Vert \tilde v_{l+1} \Vert$ (see the Eq. above Eq. (4)) and $h_l^*$ should be quantities known before the Schur decomposition. Therefore, I did not use any transformed quantities. Regarding the association of $\rho_j$, the Ritz values and eigenvalues obtained from the Schur decomposition: I thought they were roughly the same; I do not know to what extend that is actually the case.

More generally, I think we may also drop this more complex approach and may just use the plain residuals. This worked for my earlier testings and I simply adjusted this function to be more in line with the paper.

src/eigsolve/biarnoldi.jl
Jutho: The way this works is that, if the vectors and the linear map are real, the whole computation will happen in real arithmetic except for the final step. It is only in the real arithmetic case, that the Schur factorisation will return a matrix that is not strictly uppertriangular, but quasi-uppertriangular, i.e. where complex eigenvalues (which necessarily appear in complex conjugate pairs) appear in 2x2 diagonal blocks. So the real Schur decomposition returns a matrix that is block uppertriangular with only 1x1 and 2x2 diagonal blocks.

If the computation happens in complex arithmetic, H and K will be strictly upper triangular. Even if everything was real but it was manually converted to complex, still the Schur decomposition result will be different, i.e.

`complex.(schur(some_real_matrix)) != schur(complex(some_real_matrix))`

So when working in complex arithmetic, the problem will never pose itself. (Of course, in that case, you might be splitting in between a pair of complex conjugate eigenvalues in some arbitrary way, but that is not something the algorithm can control).

src/eigsolve/biarnoldi.jl
Jutho: I am certainly perfectly fine with operating solely on the Schur vectors, which is what I also do in the `Arnoldi` case. Am I correct about the fact that you rather want to use something like `Z' * M * Q` instead of just `M` for the denominators of `xh` and `xk`. I am happy to make these changes myself, but I did not want to do it before consulting you.

Jutho: I've read your comment once more. The Ritz values (eigenvalues of H and K) are indeed obtained from the Schur decomposition. But it is about the corresponding "Ritz vector", i.e. the vector c_j. That is the corresponding eigenvector of H, not the Schur vector. But also for single-sided Arnoldi I only use the Schur vector, although I then recompute the norm of the residual for the eigenvectors in the final `eigsolve` routine. I will implement a similar strategy here and then get back to you for comments.

src/eigsolve/biarnoldi.jl
tipfom: That's really nice to know! I think my observation regarding the "splitting" of eigenvalue pairs boils down to what you mentioned and was only a symptom. Hence, your suggestion works as well and I'll change the corresponding line.

Jutho:
My apologies for the long silence, I was on holiday last week and didn't really get to continue on this. I will try to finish my review / changes from my side this week.

Jutho:
Looking at this once again, I am making some more changes locally and will push them asap. However, I have two more questions:

1) One of my comments regarding the residual computation was that I think for kappa_j you need not simply M[converged+1,converged+1] but rather `Z[:,converged+1]'*M*Q[:,converged+1]`.

2) Can you elucidate what the problem is with the eigenvalues of H and K not being matched up perfectly? Is this only about complex conjugate eigenvalue pairs in the case of using real arithmetic? Or is it more general? You know only fix this in the final step, where you compute the eigenvectors, but I would like for this to be true during the Schur process. I am not sure if it is strictly necessary, but I could foresee problems, where e.g. in deciding on the `keep` variable, it might be that both H and K have 2 x 2 blocks that misalign, so that no keep can be chosen that doesn't cut through blocks in either of those two matrices.

Jutho:
Actually, looking more in the computation of the convergence criterion, I don't understand why the authors want to use a relative precision where they divide by rho_j, being the eigenvalue. This doesn't make sense in my opinion. It can be perfectly valid to target eigenvalues that happen to be close to zero, as long as they are extremal. Furthermore, adding identity to an operator changes nothing to the Krylov subspace and thus to the method, but arbitrarily shifts eigenvalues and thus the meaning of relative errors. If you agree with this, I would rather opt for a convergence criterion that only uses absolute norms of residuals in comparison to `tol`.

tipfom:
> 1. One of my comments regarding the residual computation was that I think for kappa_j you need not simply M[converged+1,converged+1] but rather `Z[:,converged+1]'*M*Q[:,converged+1]`.

Yes, that is correct. I'm however fine with dropping this prefactor alltogether as you suggested in your other comment. :)

> 2. Can you elucidate what the problem is with the eigenvalues of H and K not being matched up perfectly? Is this only about complex conjugate eigenvalue pairs in the case of using real arithmetic? Or is it more general? You know only fix this in the final step, where you compute the eigenvectors, but I would like for this to be true during the Schur process. I am not sure if it is strictly necessary, but I could foresee problems, where e.g. in deciding on the `keep` variable, it might be that both H and K have 2 x 2 blocks that misalign, so that no keep can be chosen that doesn't cut through blocks in either of those two matrices.

I've only encountered this problem when working with (mathematically) real matrices, i.e., independent of the arithmetic type (complex or real), but those who have the property that both $\lambda$ and $\lambda^\star$ are eigenvalues. Then it may happen that when sorting for smallest real the eigenvalues of $H$ and $K$ are $[\ldots, \lambda, \lambda^\star, \ldots]$ and $[\ldots, \lambda^\star, \lambda, \ldots]$ respectively. To prevent this, I implented this sort-and-match algorithm. For the `keep` part, I thought that if we put the keep threshold in between the two, i.e., keep $\lambda$ in $H$ and $\lambda^\star$ in $K$, the two of them would technically be the same as unconverged eigenvectors.

Thats as much as I know. Sadly, I did not find any information on this in the paper. When I tried enforcing this explicit ordering in the Krylov decomposition by adjusting `pK = sortperm(valuesK; by=by ∘ conj, rev=rev)` to also match to `pH`, I encountered the `cannot split 2x2 blocks when permuting schur decomposition` when calling `permuteschur!` and therefore dropped this approach.

Jutho:
Ok thanks for the comments; that is very helpful. When working with real arithmetic, a valid `which` argument should anyway be such that it doesn't distinguish between $$\lambda$$ and $$\lambda\^ast$$ (even though that is hard to check in full generality if it is provided as a general `EigSorter.by`. The complex conjugate pairs of eigenvalues are always returned by Lapack in fixed order (first the one with negative real part, then the one with positive real part). But indeed, since the eigenvalues of H and K are related by complex conjugation, it does mean that complex pairs of eigenvalues are actually swapped between H and K. But then it is in a fully deterministic manner, and we don't need complicated matching logic I believe.

tipfom:
Yes. Exactly due to this fact, the matching logic currently boils down to a linear loop which also normalizes the vectors. I'm also very open to simplifying it more :)

tipfom:
I've now gotten around to using this method some more and implementing non-Hermitian DMRG using it. In this context, I that the concerns raised by @lkdvos about the precision used in isapprox is very important, i.e., for early sweeps the left/right eigenvalues may not match perfectly as is the case with the more constructed examples in the test suite. Currently, the method should at least (i) allow the required precision as a parameter and (ii) not return an error if the algorithm does not converge. Have you @Jutho implemented a different way to return the "correctly" matched eigenvalue pairs? Otherwise, I will think some more about it and commit a suggested fix.

tipfom:
I did some more digging and in SLEPc, they've also implemented the [algorithm](https://github.com/slepc/slepc/blob/main/src/eps/impls/krylov/krylovschur/ks-twosided.c). However, they also need to run a [match loop](https://slepc.upv.es/documentation/current/src/sys/classes/ds/impls/nhepts/dsnhepts.c.html#DSSort_NHEPTS) for the eigenvalues to be sorted properly when using the same LAPACK functionality, cf. [here](https://slepc.upv.es/documentation/current/docs/manualpages/DS/DSNHEPTS.html#DSNHEPTS).

Jutho:
My apologies for the silence, I was unavailable all of last week. I will try to pick it up again this week and finally push this over the finish line.

tipfom:
Is there something else I can contribute to help get this finished @Jutho ? 😊 
I tried to come up with a more elegant solution to the eigenvalue sorting; however, none of them really yielded satisfying results. There are some improvements to the current implementation, i.e., changing the comparison for `Complex` values to only compare the norm difference such that `1+1e-19im` and `1+1e-16im` would still compare be the same which they currently do not because `isapprox` is applied to both the real and the imaginary part. This is however only a minor improvement to the current implementation and does not fundamentally fix the issue.

Jutho:
My apologies for the silence here; I got a bit distracted by other code reviews. This week I am also not actively at work very much, but I will still try to find some time. If not, then definitely early next week I will make time for finishing this.

tipfom:
I've now updated the residual norm convention to be with `rV` and `rW`.

src/eigsolve/biarnoldi.jl
Jutho: Is it ok that I work on the code for a bit now and you don't push further commits at this point? I think actually extracting the norms `βv` and `βw` like it was before the last commits is actually fine; I do like it when implementations follow the reference paper. I am just adding some comments to clarify things, and I was mostly concerned that `xh` requires the extra factor `βrV`. If it is ok I will do a force push to overwite the last commits and add some changes of my own.

src/eigsolve/arnoldi.jl
```diff
@@ -475,10 +475,5 @@ function _restorearnoldiform!(U, H, f, keep)
         rmul!(view(H, 1:j, :), h')
         rmul!(U, h')
     end
-    copyto!(Hcompact, H) # copy back into compact Hessenberg form
-
-    # Update basis B by applying U
-    basistransform!(B, view(U, :, 1:keep))
-    B[keep + 1] = scale!!(r, 1 / βr)
```
Jutho: Oh right I forgot to remove these. I had so far only tested `_bischursolve` with `maxiter=1` before pushing, so I hadn't tested this part yet indeed. Thanks for fixing.

tipfom:
Hi @Jutho, I like this new approach with inverting `M` much better! This now also allows the algorithm to be used to get the current best estimate of both left and right eigenvector without full convergence, which is really neat. Thanks! I've pushed the changes which I needed to get it to run on my local machine.

The only major change is that at least for me working with `conj(Q)` and `conj(Z)` and `h'` and `k' yielded wrong results.
Can you maybe confirm that this happens for you as well, e.g., by running
```julia
  using LinearAlgebra
  using KrylovKit
  using Random
  
  function main()
      Random.seed!(111111)
  
      A = rand(ComplexF64, 1000, 1000)
      x0 = rand(eltype(A), 1000)
  
      fAr = x -> A * x 
      fAl = x -> Adjoint(A) * x
  
      vals, (V, W), info = bieigsolve((fAr, fAl), x0, x0, 3, :SR, BiArnoldi(krylovdim=10, maxiter=1000, tol=1e-10))
  
      @show eigvals(A)[1:5]
  
      for i in eachindex(vals)
          @show vals[i]
          @show KrylovKit.inner(W[i], fAr(V[i]))
      end
  end
  
  main()
```

src/eigsolve/biarnoldi.jl
```diff
@@ -261,8 +261,8 @@ function _bischursolve(f, v₀, w₀, howmany::Int, which::Selector, alg::BiArno
             VQᴴv = -view(Q, :, 1:keep)' * M⁻¹Wᴴv
             WZᴴw = -view(Z, :, 1:keep)' * M⁻ᴴVᴴw
 
-            H[1:keep, 1:keep] += VQᴴv * h[1:keep]' # Ĥ = Sₖₖ + (V * Qₖ)' * ṽℓ₊₁ * h̃ₖ'
-            K[1:keep, 1:keep] += WZᴴw * k[1:keep]' # K̂ = Tₖₖ + (W * Zₖ)' * w̃ℓ₊₁ * k̃ₖ'
+            H[1:keep, 1:keep] += VQᴴv * transpose(h[1:keep]) # Ĥ = Sₖₖ + (V * Qₖ)' * ṽℓ₊₁ * h̃ₖ'
+            K[1:keep, 1:keep] += WZᴴw * transpose(k[1:keep]) # K̂ = Tₖₖ + (W * Zₖ)' * w̃ℓ₊₁ * k̃ₖ'
```
Jutho, line 265: Are you sure about those changes? It depends on whether you want `h` and `k` to be the transpose of the row vector in the Krylov factorization "A * V = V * H + r * (row vector)" or the adjoint, but I would prefer the adjoint, which is also how it is in the paper if I recall correctly. I think if you want to change it back to transpose, there are other places where it needs to be changed as well, like where the residuals of the eigenvectors are computed.

Jutho: Ok I'll take a further look later today.

Jutho: I decided to change the Arnoldi implementation to always represent the row vector as the (Hermitian) `adjoint` of a column vector, as this is also how it was written in the documentation of `KrylovFactorization` etc.

tipfom:
@Jutho, I've now added the tests. Sorry for the long delay :)

tipfom:
Yes, I removed it because as of now the option was not working as intended, i.e., the unit tests failed with `eager=true`. However, I did not investigate this further and don't see an immediate reason why this would necessarily be the case. Is this something we aspire to include? If so, I could spend some time tomorrow to figure out if it is an algorithmic constraint or just some operation I missed to apply when using the `eager` option.

Jutho:
@tipfom , I finally made some time to look at this again, and would like to get this merged before tagging a new release of KrylovKit. However, currently, I get completely wrong results as soon as `maxiter > 1`, i.e. as soon as a restart takes place. Was this working before for you and did I manage to break something?

I did rebase this PR on the current master. Apparently I also had some uncommited changes about the role of the `h` and `k` vectors, so maybe this is what messed something up?

tipfom:
The `eager` option is now fixed. You may want to look at the commit 467ca6f in which I changed the residual calculation because `Base.Fix1` was causing `@constinferred`  to fail sometimes.

Jutho:
> The `eager` option is now fixed. You may want to look at the commit [467ca6f](https://github.com/Jutho/KrylovKit.jl/commit/467ca6f17da771d00e0525989ed9d67cd4b9095f) in which I changed the residual calculation because `Base.Fix1` was causing `@constinferred` to fail sometimes.

Yes, I also noticed this last night but don't understand why. The `Base.Fix1` is supposed to be the more robust solution as it doesn't create a closure over the first variable, being `hᴴVS`. Also, it seemed very undeterministic but maybe I missed something.

Issue #87, opened 2024-05-28
Feature request: block lanczos for solving degenerate ground states


GiggleLiu:
I used the following code for demonstrating the 4-fold degeneracy of toric code model, without success. I leave an issue here so that someone (including me) could inspect it in the future.

```julia
using Yao

function toric_code_strings(m::Int, n::Int)
	li = LinearIndices((m, n))
	bottom(i, j) = li[mod1(i, m), mod1(j, n)] + m * n
	right(i, j) = li[mod1(i, m), mod1(j, n)]
	xstrings = Vector{Int}[]
	zstrings = Vector{Int}[]
	for i=1:m, j=1:n
		# face center
		push!(xstrings, [bottom(i, j-1), right(i, j), bottom(i, j), right(i-1, j)])
		# cross
		push!(zstrings, [right(i, j), bottom(i, j), right(i, j+1), bottom(i+1, j)])
	end
	return xstrings, zstrings
end

function toric_code_hamiltonian(m::Int, n::Int)
	xstrings, zstrings = toric_code_strings(m, n)
	sum([kron(2m*n, [i=>X for i in xs]...) for xs in xstrings[1:end-1]]) + sum([kron(2m*n, [i=>Z for i in zs]...) for zs in zstrings[1:end-1]])
end

h = toric_code_hamiltonian(3, 3)

using KrylovKit
eigsolve(-mat(h), 10, :SR)
```

I got the following output:
```
julia> eigsolve(-mat(h), 10, :SR)[1]
17-element Vector{Float64}:
 -15.999999999999964
 -13.999999999999982
 -11.999999999999954
  -9.999999999999996
  -8.000000000000004
  -5.999999999999954
  -3.9999999999999893
  -2.0000000000000675
  -2.1316282072803006e-14
   2.0000000000000604
   4.000000000000041
   5.999999999999986
   7.999999999999975
   9.999999999999996
  11.999999999999982
  14.000000000000005
  16.000000000000007
```

I hope block Lanczos could be supported in the future so that degeneracies could be resolved correctly. For someone who might be interested in implementing this, please check:
- Matrix Computation, by Golub, Sec. 10.3.6

tibidabo:
Hello why is issue this closed? I would also be interested in a Block Lanczos version.

hz-xiaxz:
Hello, for your case, I tried [ArnoldiMethod.jl](https://github.com/JuliaLinearAlgebra/ArnoldiMethod.jl) and get the output
[...]

PR #135, opened 2025-07-24
Improve field types of blocklanczos methods

gdalle:


Jutho:
Thanks for these fixes. The `BlockLanczos` method was implemented by a contributor, and clearly I have overlooked these abstract fields when reviewing the PR before merging.

src/algorithms.jl
```diff
@@ -148,12 +148,12 @@ Use `Arnoldi` for non-symmetric or non-Hermitian linear operators.
...
 See also: [Factorization types](@ref), [`eigsolve`](@ref), [`Arnoldi`](@ref), [`Orthogonalizer`](@ref)
 """
-struct BlockLanczos{O<:Orthogonalizer,S<:Real} <: KrylovAlgorithm
+struct BlockLanczos{O<:Orthogonalizer,S<:Real,S2<:Real} <: KrylovAlgorithm
     orth::O
     krylovdim::Int
     maxiter::Int
     tol::S
-    qr_tol::Real
+    qr_tol::S2
```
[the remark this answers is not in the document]
Jutho: Yes I agree with @lkdvos and would actually like that more. There can then be a constructor that catches different types and just promotes them to a common type.

lkdvos:
I spent some time trying to improve a bit the types in general.
As discussed, I made sure to promote some of the types to reduce the number of type parameters (and different methods to compile).
Additionally, I hardcoded the matrix type of the block factorization to simply be `Matrix{S}`. While this could probably benefit from a more abstract type (thinking of CuArray mostly), the current interface actually would never expose that, and the remainder of the code is not written in a way that would be compatible with GPU implementations anyways.

I refactored a bunch of the `Block` methods as well, since I don't think using `Block{S}(vec) = Block{T,S}(vec)` is a pattern that is good practice.
In general though, going through that code in a little more detail made me realize that it could really benefit from a larger clean-up and refactor, as there are quite a few somewhat questionable choices still in there.

If the tests now pass, I think I'm okay with merging as is though, since I don't have the time to really go through it.

src/factorizations/blocklanczos.jl
```diff
@@ -8,57 +8,56 @@ while `S` represents the type of inner products between vectors. To create an in
...
-    M = Matrix{S}(undef, length(B₁.vec), length(B₂.vec))
-    @inbounds for j in 1:length(B₂)
-        yj = B₂[j]
-        for i in 1:length(B₁)
-            M[i, j] = inner(B₁[i], yj)
-        end
+
+apply(f, block::Block) = Block(map(Base.Fix1(apply, f), block.vec))
+
+function VectorInterface.inner(B₁::Block{T}, B₂::Block{T}) where {T}
```
Jutho, line 36: Since `Block` is really a special type that is not supposed to represent a single vector, I would prefer to not make this another method definition of `VectorInterface.inner` (which, I would argue, returns a scalar value as part of the definition), and thus keep the `block_inner` name. There are some other `block_...` methods as well. I just noticed there is also a `blocklanczosrecurrence` which should have been `block_lanczosrecurrence` for consistency.

src/factorizations/blocklanczos.jl
```diff
@@ -288,14 +280,19 @@ and `r` is the numerical rank of the input block. The matrix represents the uppe
...
     n = length(block)
     rank_shrink = false
     idx = trues(n)
-    R = zeros(S, n, n)
+    r₁₁ = inner(block[1], block[1])
+    R = zeros(typeof(r₁₁), n, n)
     @inbounds for j in 1:n
         for i in 1:(j - 1)
-            R[i, j] = inner(block[i], block[j])
+            if j == i == 1
```
Jutho, line 291: This cannot happen here since `i in 1:(j-1)`

src/factorizations/blocklanczos.jl
```diff
@@ -126,38 +124,36 @@ Here, [`initialize(::KrylovIterator)`](@ref) produces the first Krylov factoriza
...
-                    verbosity::Int=KrylovDefaults.verbosity[]) where {F,T,S}
+function initialize(iter::BlockLanczosIterator;
+                    verbosity::Int=KrylovDefaults.verbosity[])
     X₀ = iter.x₀
     maxdim = iter.maxdim
     A = iter.operator
-    BTD = zeros(S, maxdim, maxdim)
 
     # Orthogonalization of the initial block
     X₁ = copy(X₀)
```
Jutho, line 159: Rather than a plain copy, here we should make sure that `X₁` is such that it supports in place `block_qr!`, i.e. it contains vectors with floating point scalar types. I will think some more about the best/nicest way to accomplish this.

Jutho:
Thanks @lkdvos for carefully going through the `BlockLanczos` implementation. Even though this is a bit beyond the initial intent of this PR, I think we can finalize this rather quickly. I made a first batch of comments, but will look at it again after these have been addressed. It might also be worth to have a review or hear possible comments by @yuiyuiui.

src/factorizations/blocklanczos.jl
```diff
@@ -8,57 +8,56 @@ while `S` represents the type of inner products between vectors. To create an in
...
-        for i in 1:length(B₁)
-            M[i, j] = inner(B₁[i], yj)
-        end
+
+apply(f, block::Block) = Block(map(Base.Fix1(apply, f), block.vec))
+
+function VectorInterface.inner(B₁::Block{T}, B₂::Block{T}) where {T}
+    m₁₁ = inner(B₁[1], B₂[1])
+    M = Matrix{typeof(m₁₁)}(undef, length(B₁), length(B₂))
+    @inbounds M[1, 1] = m₁₁
```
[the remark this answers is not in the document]
lkdvos: Why do we need this to be a floating point type? If `inner` is type-stable, this would just be a matrix of inner products, and then possibly we need to promote this to a floating point type later when taking square roots but that should be fine?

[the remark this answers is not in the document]
lkdvos: and if we do, can I just use `M = Matrix{float(typeof(m₁₁))}`?

src/factorizations/blocklanczos.jl
```diff
@@ -255,8 +253,7 @@ Specifically, it modifies each vector `R[i]` by projecting out its components al
 
 Here,`⟨·,·⟩` denotes the inner product. The function assumes that `V` is already orthonormal.
 """
-function block_reorthogonalize!(R::Block{T,S},
-                                V::OrthonormalBasis{T}) where {T,S}
+function block_reorthogonalize!(R::Block{T}, V::OrthonormalBasis{T}) where {T}
```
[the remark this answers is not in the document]
yuiyuiui: Thank you for taking the time to review the code I wrote.

Regarding the choice of orthogonalization method: I originally developed `BlockLanczos` to address the [issue 87](https://github.com/Jutho/KrylovKit.jl/issues/87) concerning repeated (degenerate) eigenvalues. Among the methods I implemented — `CGS` (Classical Gram-Schmidt), `CGS2`, `MGS`, `MGS2`, and Householder — only `MGS2` was able to reliably resolve the four-fold eigenvalue case required by the issue accurately. That's why I chose to implement only `MGS2`.

That said, I was still quite new to open-source development at the time, and I didn’t include performance benchmarks of other orthogonalization methods, which I probably should have.

I do agree that `MGSIR` should be implemented as well, and I plan to work on this in September.

src/factorizations/blocklanczos.jl
```diff
@@ -289,18 +289,14 @@ function block_qr!(block::Block, tol::Real)
     n = length(block)
     rank_shrink = false
     idx = trues(n)
-    r₁₁ = inner(block[1], block[1])
+    r₁₁ = norm(block[1])
```
yuiyuiui, line 292: I would like to ask about the reason for including these two lines of code.
```julia
r₁₁ = norm(block[1])
β = j == 1 ? r₁₁ : norm(block[j])
```
It seems to me that if they were removed, neither the correctness nor the performance would be noticeably affected. Could you please clarify?

lkdvos: The problem is that we need to allocate `R` with the correct `eltype` before we can fill this in. This can be achieved by computing a single entry and then making the entire matrix, but I then want to re-use the fact that I already computed this single entry.

An alternative would be to depend on type inference to check what the output type would be, but this has its own set of problems so I'd like to avoid that.

Another alternative is to simply always assume that `typeof(inner(::T, ::T)) = float(scalartype(T))`, but again that seems like it might break some things.

Looking at the current implementation, I'm also no longer certain that this really is correct, given that I guess `inner` might return a complex number while `norm` shouldn't do that?

src/algorithms.jl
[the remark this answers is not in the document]
lkdvos: In the current change this would still work though: the implementation will automatically promote both tolerances to a common type, and simply compare to that value. (in your example, you'd simply get `Float64(eps(Float32))` as the `tol`, which would for all intents and purposes yield the same results)

src/factorizations/blocklanczos.jl
yuiyuiui: Thank you, I understand your point now. I noticed that you removed the `S` parameter representing the scalar field of the inner product space from the `Block` type, so using `r₁₁` to determine the scalar type makes sense in this context.

If I may, I’d like to briefly review the mathematical concept of an inner product space—although I’m sure you are already very familiar with it.

An inner product space is usually denoted as
$((X, K),~ (\cdot,\cdot))$
where $X$ is the set of vectors, and $K$ is a field lying between $\mathbb{Q}$ and $\mathbb{C}$. That is, $(X, K)$ is a vector space and $(\cdot,\cdot)$ is the inner product map:
$(\cdot,\cdot): ~X \times X \rightarrow K$

From this viewpoint:

1. Since $K$ lies between $\mathbb{R}$ and $\mathbb{C}$, we don’t need to worry about `inner` returning a complex number—it is natural to assert that `typeof(inner(u, v)) = K`. (Though I understand this is not strictly correct in a formal mathematical sense.)
2. Since $K$ might be something like 

$$(a + b\sqrt{p} : a, b \in \mathbb{Q} ),~~\ p \in \mathbb{Z}^+ $$

and now `r₁₁` might not be a subtype of `Number`. It could, for example, be a user-defined type like:

   ```julia
   struct SqrtP{I <: Int}
       a1::I
       a2::I
       b1::I
       b2::I
   end
   ```

   So assuming `float` in `float(scalartype(T))` may not always be appropriate.

Additionally, the `vec` field in `InnerProductVec` does not necessarily have to be of type `Vector`; it could also be a custom type. Thus, `scalartype(T)` may not apply reliably to all `T`.

Overall, I believe that using the type of `r₁₁` is the correct approach.

src/factorizations/blocklanczos.jl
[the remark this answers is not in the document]
Jutho: You are right, for the `block_inner` matrix, the floating point promotion is indeed not necessary. It is more relevant for the `block_qr!` as remarked in my other comment.

src/factorizations/blocklanczos.jl
Jutho: @yuiyuiui , I don't think it makes much sense to consider inner product spaces where the field is not $\mathbb{R}$ or $\mathbb{C}$. Indeed, the Wikipedia definition includes real or complex as part of the definition:
https://en.wikipedia.org/wiki/Normed_vector_space
https://en.wikipedia.org/wiki/Inner_product_space

You need to be able to take square roots in order to go from the inner product to the norm.

Also, assuming `scalartype(T)` or `float(scalartype(T))` as the type of the inner product would be a bad idea, as the `RealVec` is a way to wrap arbitrary (typically complex) vectors in such a way that the vector space is considered a real vector space, with a real inner product (but represented using a complex `scalartype`).

Jutho: Also, indeed, @lkdvos, the element type of `R` should really be determined by an `inner` call and not by `norm` call.

lkdvos: I've updated this now, but I also got confused about the values in the matrix. Is it correct that this had to be the norm? Previously, we had more or less `R[i,j] = i == j ? norm(block[i]) : inner(block[i], block[j])`, which seems like these quantities scale differently. Am I seeing ghosts here or was this a small mistake?

Jutho: No that is correct, this is because you rescale the final vector to have norm 1. A single column from the QR decomposition effectively reads

```
v = norm(orthogonal_remainder) * normalized_orthogonal_remainder + sum( inner(ei, v) * ei for ei in previous_orthonormalized_vectors)
```

or thus

```
V[:, i ] = Q[:, i] * R[i,i] + sum(j->Q[:,j] * R[j, i], j=1:i-1)
```

and thus `R[i,i]` is the norm of whatever remains after you subtracted all the orthogonal projections onto the previous vectors.

lkdvos: okay, then I think the current implementation should be good, I'll resolve this conversation

Jutho:
Ok, I've made some changes; if CI is succeeds, then I am happy/done with my changes, and this is perhaps ready for a (hopefully final?) review.

src/factorizations/blocklanczos.jl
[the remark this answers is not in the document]
lkdvos: Great, I'll resolve this discussion for now, happy to re discuss when you have more time, I don't think this is super urgent.

PR #140, opened 2025-10-09
mark scalartype determining functions as non-differentiable

### LINKED ISSUE #139: AD error for v0.10.1
I tried the following code
```Julia
using LinearAlgebra, KrylovKit, Zygote

f(x) = begin
    M = reshape(x, 3, 3)
    M = (M + M') / 2
    vals, vecs, info = eigsolve(M, ones(3), 1, :LM)
    return vals[1]
end

x0 = randn(9)
value, grad = Zygote.withgradient(f, x0)
```
For v0.10.1, `ERROR: Tuple field type cannot be Union{}` is reported. For v0.9.5, it passed.

lkdvos:
Closes #139


Jutho:
Thanks. Clearly there is no JET in the KrylovKit CI 😄 . Maybe something to add, once it is functional again.

PR #141, opened 2025-10-10
Update format action

lkdvos:
This should fix the failing actions on pushes to main, and only run the format test on PRs. (doesn't require a new release, just might be nice to have this fixed for the future)


Jutho:
Thanks.

Issue #143, opened 2025-12-21
`BlockLanczos` on a `nxn` matrix returns more than `n` eigenvalues


mxhbl:
Hi!

I just ran into a weird issue, where using `BlockLanczos` with large block sizes returns more eigenvalues than a matrix has. Moreover, all of the returned eigenvalues are not converged. I was only able to reproduce this on a few specific matrices; I shared one of these below.

```julia-repl
julia> using KrylovKit, LinearAlgebra

# A is a 71 x 71 matrix, see below for value

julia> eigvals(A)[1:4]
4-element Vector{Float64}:
 -1.2525363565095453e-11
 -7.872263865818092e-13
  8.139934675911611e-14
  1.0000000000170322

julia> λs, vs, info = eigsolve(A, Block([randn(71) for _ in 1:20]), 4, :SR, BlockLanczos(tol=1e-8, verbosity=2));
┌ Info: BlockLanczos eigsolve finished after 80 iterations:
│ * 80 eigenvalues converged
[...]

PR #145, opened 2025-12-24
Eliminate draft subspaces in block_qr!

yuiyuiui:
I have fixed the issue of subspace drift described in (https://github.com/Jutho/KrylovKit.jl/issues/143). The cause of the problem, my solution, and the results can be found in (https://github.com/Jutho/KrylovKit.jl/issues/143#issuecomment-3687943668). Specifically, I used DGKS reorthogonalization to improve the accuracy of `block_qr!`, ensuring that the vectors within the `Q` produced by `block_qr!` are strictly orthogonal (“Reorthogonalization and Stable Algorithms for Updating the Gram–Schmidt QR Factorization”, "Numerical Methods for Least Squares Problems"). Then, by projection, I removed components along `V` from the `Q` obtained in `block_qr!` to guarantee that `Q` is orthogonal to `V`. Afterwards, I applied `block_qr!` again to `Q` to ensure orthogonality among its vectors. It can be proven that this second `block_qr!` preserves the subspace at machine precision, resulting in a block that lies strictly in the orthogonal complement of `V` and whose vectors are strictly mutually orthogonal.

Due to updates in JuliaFormatter.jl, I had to reformat a large portion of the code. The actual code changes I made are only in "src/factorizations/blocklanczos.jl". Including function `initialize`, `expand!` and `block_qr!`.

In addition, the parts of the CI that are failing seem unrelated to my changes; they were failing before my modifications.


lkdvos:
Hi @yuiyuiui, thanks for the quick reply and work!

We have switched over to the [Runic.jl formatter](https://github.com/fredrikekre/Runic.jl), which is why you are seeing such large changes when still trying to use JuliaFormatter.jl, and our format CI is not passing.
Is there any chance you could undo these changes and use the runic formatter instead?
It is a bit hard to see what is going on like this.

yuiyuiui:
Hi @lkdvos , thanks for the clarification.

The large diff was caused by JuliaFormatter, and unfortunately those changes cannot be cleanly reverted using Runic directly. I’m still not very experienced with Git version control, so instead of trying to partially undo the formatting commits, I opened a new PR that contains only my actual code changes.

This should hopefully make it much easier to review what was modified. Sorry for the inconvenience caused by the formatter mismatch, and thanks again for your patience and guidance.

PR #146, opened 2025-12-24
reformat and revise subspace drift

yuiyuiui:
I have fixed the issue of subspace drift described in (https://github.com/Jutho/KrylovKit.jl/issues/143). The cause of the problem, my solution, and the results can be found in (https://github.com/Jutho/KrylovKit.jl/issues/143#issuecomment-3687943668). Specifically, I used DGKS reorthogonalization to improve the accuracy of `block_qr!`, ensuring that the vectors within the `Q` produced by `block_qr!` are strictly orthogonal (“Reorthogonalization and Stable Algorithms for Updating the Gram–Schmidt QR Factorization”, "Numerical Methods for Least Squares Problems"). Then, by projection, I removed components along `V` from the `Q` obtained in `block_qr!` to guarantee that `Q` is orthogonal to `V`. Afterwards, I applied `block_qr!` again to `Q` to ensure orthogonality among its vectors. It can be proven that this second `block_qr!` preserves the subspace at machine precision, resulting in a block that lies strictly in the orthogonal complement of `V` and whose vectors are strictly mutually orthogonal.


Jutho:
I had forgotten about this PR. Feel free to send me reminders occasionally. I won't have time this week but will try to make time next week.

yuiyuiui:
@Jutho No problem! I've added tests for issue #143 and corrected the wording from 'draft' back to 'drift'. Looking forward to your feedback.

yuiyuiui:
Hi @Jutho  , just a friendly reminder regarding my PR when you have a moment. Please let me know if there’s anything I should update or improve. Thanks!

src/factorizations/blocklanczos.jl
```diff
@@ -173,7 +173,8 @@ function initialize(
     X₁ = Block(map(Base.Fix2(scale, one(α)), X₀.vec))
 
     # Orthogonalization of the initial block
-    _, good_idx = block_qr!(X₁, iter.qr_tol)
+    _, good_idx, is_drift = block_qr!(X₁, iter.qr_tol)
+    is_drift && @warn "Block size may be more than operator's rank. Please decrease Krylovdim or increase qr_tol."
```
Jutho:
```suggestion
    is_drift && @warn "Block size may be more than operator's rank. Please decrease `krylovdim` or increase `qr_tol`."
```

Jutho: Although I am confused by the suggestion to reduce `krylovdim`. Is it not rather the number of vectors in the initial `x₀::Block` that you want to see reduced? Most other methods in KrylovKit.jl don't error on a `krylovdim` setting that would be too large, and would just detect this and gracefully continue with the smaller Krylov subspace, i.e. the `krylovdim` value is interpreted as maximal Krylov subspace dimension.

src/factorizations/blocklanczos.jl
```diff
@@ -312,28 +322,42 @@ function block_qr!(block::Block, tol::Real)
...
+            R[j, j] = β
+            block[j] = scale!!(block[j], 1 / β)
+            # DGKS reorthogonalization
+            if β < 100 * tol
+                is_drift = true
+                for i in 1:(j - 1)
+                    δ = inner(block[i], block[j])
+                    R[i, j] += δ
+                    block[j] = add!!(block[j], block[i], -δ)
+                end
```
Jutho: Are you sure this reorthogonalization is correct. You have already rescaled `block[j]` on line 341, so I think this needs to be reflected here. Typically, when reorthogonalizing, one doesn't already rescale the vector being orthogonalized until the very end. Furthermore, the criterium for when to reorthogonalize is based on a relative change in norm before and after the orthogonalization step. See e.g. here
https://github.com/Jutho/KrylovKit.jl/blob/03fa1630aaa269e1bb1f46c11d46aecbbebbd934/src/orthonormal.jl#L440

yuiyuiui:
Thank you for your comments. I would like to clarify the rationale behind my implementation and other code changes in the following three points.

First, regarding the decision to rescale `block[j]` prior to the second orthogonalization: I believe this approach offers better numerical precision compared to the typical reorthogonalization workflow. While the two methods are theoretically equivalent, a potential issue arises when `block[j]` has an extremely small norm. In the standard approach, any tiny rounding errors in the floating-point mantissa are amplified by a massive factor during the final rescaling step. By performing an intermediate scaling before the second orthogonalization, we ensure that `block[j]` remains $O(1)$ during the final normalization, which significantly mitigates the amplification of these errors. Although I haven't yet conducted a formal rigorous study on this, my preliminary mathematical estimates suggest that the improvement in error is proportional to the magnitude of the largest element in the output vector. As a numerical demonstration, in the provided example, the ratio of the errors between the two methods is approximately $O(10^{-d/2})$ with the vector length being $10^d$, which aligns with the observed results. Therefore, I believe incorporating this intermediate scaling step can enhance the robustness of the reorthogonalization process.

```julia
using Random, LinearAlgebra, Plots

err_rate = Float64[]

maxd = 8
Random.seed!(123)

for d in 1:maxd

    n = 10^d

    Q = qr(randn(n, 2)).Q

    v = (Q[:, 1] + Q[:, 2]) * 1e-12

    u = Q[:, 2]

    v1 = v - dot(v, u) * u
    normalize!(v1)
    norm(v1)

    v2 = normalize(v1)
    v2 = v2 - dot(v2, u) * u
    normalize!(v2)
    norm(v2)

    err1 = abs(dot(v1, u))

    err2 = abs(dot(v2, u))

    push!(err_rate, err2 / err1)
end

plot(1:maxd, err_rate, xlabel="log10(n)", label="err ratio", yscale=:log10)
plot!(1:maxd, 10 .^ ( - (1:maxd) / 2), label="10^(-n/2)")
```

e598c76ebf05795b46da2c55e1a42cae

yuiyuiui:
Second, I would like to explain why I did not use the relative change in the vector norm as the criterion for triggering reorthogonalization. There are two main considerations:

1. Consistency with the global tolerance: We already have a unified scale, `qr_tol`, to determine whether a vector is numerically zero during the QR process. If we were to use a relative threshold for subspace drift, we might encounter cases where the drift-detection threshold is actually smaller than `qr_tol`. Conceptually, this creates ambiguity. More importantly, the primary cause of subspace drift is often when $\beta$ is only slightly larger than `qr_tol`. If the drift threshold is set lower than `qr_tol`, these critical cases might be bypassed. I have not found any theoretical guarantee that severe subspace drift would not occur under such circumstances.

2. Computational efficiency in high-dimensional spaces: Classical reorthogonalization theory suggests that if $\beta < \|v\|/\sqrt{2}$ after orthogonalization, a second pass is required. However, in high-dimensional spaces, for a given low-dimensional subspace $X$, the measure of unit vectors with an angle to $X$ of $\pi/4$ or less is nearly zero relative to the total area of the unit sphere. Applying such a stringent "numerical safety" standard might be overly conservative, leading to unnecessary reorthogonalization steps and wasted computational resources. I believe a more relaxed criterion is sufficient for maintaining stability without sacrificing performance.

Admittedly, I do not yet have a definitive conclusion on what the mathematically optimal criterion should be. In the current implementation, I have tentatively adopted the condition `tol < β < 100 * tol` as the trigger for "potential subspace drift requiring additional orthogonalization."

I should clarify that this threshold is based on my practical experience and numerical intuition encountered during testing, rather than a rigorous theoretical derivation. My goal was to provide a robust safeguard for these edge cases while maintaining computational efficiency.

yuiyuiui:
Third, regarding the `is_drift` flag and the associated `@warn` messages: In my latest commit d291ddf60515a2b5d5928671cc6dd4611a11f2ce, I have removed the sections that trigger a warning when `is_drift` is true.
﻿
The rationale is that our updated logic can now robustly handle cases where subspace drift occurs. I believe it is unnecessary to alert the user or prompt them to adjust their input parameters. The goal is to provide a more seamless experience where the solver remains stable even when encountering these edge cases.

yuiyuiui:
>First, regarding the decision to rescale `block[j]` prior to the second orthogonalization: I believe this approach offers better numerical precision compared to the typical reorthogonalization workflow. While the two methods are theoretically equivalent, a potential issue arises when `block[j]` has an extremely small norm. In the standard approach, any tiny rounding errors in the floating-point mantissa are amplified by a massive factor during the final rescaling step. By performing an intermediate scaling before the second orthogonalization, we ensure that `block[j]` remains $O(1)$ during the final normalization, which significantly mitigates the amplification of these errors. Although I haven't yet conducted a formal rigorous study on this, my preliminary mathematical estimates suggest that the improvement in error is proportional to the magnitude of the largest element in the output vector. As a numerical demonstration, in the provided example, the ratio of the errors between the two methods is approximately $O(10^{-d/2})$ with the vector length being $10^d$, which aligns with the observed results. Therefore, I believe incorporating this intermediate scaling step can enhance the robustness of the reorthogonalization process.

Upon further reflection, I must apologize as my previous test case for reorthogonalization was not sufficiently representative.

After conducting more extensive numerical experiments, I found that while an additional `scale!!` step can slightly reduce errors in certain specific edge cases, the precision remains essentially identical to the standard approach in most scenarios. Consequently, I have concluded that the extra `scale!!` step is not strictly necessary. I have reverted this logic and implemented the standard reorthogonalization procedure in my latest commit.

yuiyuiui:
Hi @Jutho  , just a friendly reminder regarding my PR when you have a moment. Please let me know if there’s anything I should update or improve. Thanks!

src/factorizations/blocklanczos.jl
```diff
@@ -295,14 +301,16 @@ This function performs a QR factorization of a block of abstract vectors using t
 It takes as input a block of abstract vectors and a tolerance parameter, which is used to determine whether a vector is considered numerically zero.
 The operation is performed in-place, transforming the input block into a block of orthonormal vectors.
 
-The function returns a matrix of size `(r, p)` and a vector of indices goodidx. Here, `p` denotes the number of input vectors,
+The function returns a matrix of size `(r, p)`, a vector of indices goodidx and a boolean flag is_drift. Here, `p` denotes the number of input vectors,
 and `r` is the numerical rank of the input block. The matrix represents the upper-triangular factor of the QR decomposition,
 restricted to the `r` linearly independent components. The vector `goodidx` contains the indices of the non-zero
 (i.e., numerically independent) vectors in the orthonormalized block.
+If a small value of β is detected, the function will carry out an additional reorthogonalization step to further ensure the input block vectors are orthonormalized.
```
Jutho: I don't think it is mentioned anywhere what `β` is.

src/factorizations/blocklanczos.jl
```diff
@@ -312,28 +320,43 @@ function block_qr!(block::Block, tol::Real)
...
+        if β < tol
             block[j] = zerovector!!(block[j])
-            rank_shrink = true
             idx[j] = false
+            continue
+        else
+            R[j, j] = β
+            # DGKS reorthogonalization
+            if β < 100 * tol
+                is_drift = true
```
Jutho: I forgot about the precise role of `is_drift`. Do you want it `true` also if after the reorthogonalisation, `β < tol` and vector `j` is discarded? Or only if it had very small norm (and has thus been rescaled by a large vector to make it a unit vector) but it was kept?

src/factorizations/blocklanczos.jl
```diff
@@ -205,9 +205,16 @@ function expand!(
...
     bs = length(R)
     V = state.V
+    Rcopy = copy(R)
 
     # Calculate the new basis and B
-    B, good_idx = block_qr!(R, iter.qr_tol)
+    B, good_idx, is_drift = block_qr!(R, iter.qr_tol)
+    if is_drift # Prevent column subspace of R from drifting caused by an excessively small β in block_qr!
+        block_reorthogonalize!(R, V)
+        _, good_idx, is_drift = block_qr!(R, iter.qr_tol)
```
Jutho, line 214: Is this value of `is_drift` is `true`, is anything done with this? Does need to be a `while isdrift` loop?

src/factorizations/blocklanczos.jl
Jutho: Looking back on how `is_drift` is used, I think this is necessary only if vector `j` is kept, though of course it doesn't harm to do an extra unnecessary reorthogonalization step.

yuiyuiui: Yes, you are right. `is_drift` is set to `true` only when $\beta$ is not sufficiently small to be considered zero, yet small enough to potentially induce subspace drift. At this time `j` is kept.

src/factorizations/blocklanczos.jl
yuiyuiui: Yes, when `is_drift` is `true` we need to do some processing, and that is why I set the tolerance of `is_draft` just to `100 * tol` but not  `norm(block[j])/sqrt(2)` as usual reorthogonalization method does. Before `block_qr!`, the column space of `R` is orthogonal to current basis `V`. So mathematically after `block_qr!`, `R` should be orthogonal to current `V`. But subspace drift breaks that and we need to do orthogonalization to make sure `R` is orthogonal to `V`. This is computationally expensive so I set `100 * tol` to avoid frequent reorthogonalization. However, I am also not sure should I increase tolerance of drift to improve orthogonalization ? The relationship between the loss of orthogonalization and tolerance of drift scales as `err = eps(T)/tol_drift`.
```julia
using LinearAlgebra, Random, Test, Plots
Random.seed!(1234567)

err = Float64[]
tol = Float64[]

η_vec = 10.0 .^ (-4:-1:-15)

for η in η_vec
    n = 100000
    A = rand(n, 3)

    Q = qr(A).Q

    q1 = Q[:, 1]
    q2 = Q[:, 2]

    v1 = η * q2 + q1

    normalize!(v1)

    v2 = v1 - dot(v1, q1) * q1

    push!(tol, norm(v2))

    normalize!(v2)

    push!(err, abs(dot(v2, Q[:, 3])))
end

plot(tol, err; xlabel="tolerance", ylabel="error of orthogonalization",
     title="Error vs. tolerance",
     xscale=:log10, yscale=:log10)
```

80e9802d0a3c78e1ef02e102ac5be850

Although all tests pass, do you think it is necessary to increase `tol_drift`?

src/factorizations/blocklanczos.jl
yuiyuiui: Good catch, added a clarification that $\beta$ refers to the norm of the vector after the first orthogonalization.

yuiyuiui:
Hi @Jutho, just a friendly reminder regarding my PR when you have a moment. Please let me know if there’s anything I should update or improve. Thanks!

src/factorizations/blocklanczos.jl
Jutho: Well, despite the dip which makes the tolerance = 10^{-10} quite good (error of orthogonalization also below 10^{-10}, which is probably good enough), this seems to be a random noise artefact, and from the general trend it seems more like with tolerance 10^{-10}, you would get an error of orthogonalization of the order of 10^{-7} or 10^{-8}. I don't know if that is good enough.

Jutho: Also, I don't know what happened to my previous question, which clearly contains a language error, but more explicitly the question was this:

You have the first `is_drift` value at line 211, which decides whether you enter the `if is_drift` block to reorthogonalize against V. But then you also do a new `block_qr!(R, iter.qr_tol)` which yields a new `is_drift` value on line 214. Does anything need to be done if this new `is_drift` value is again true? Do you ever need to reorthogonalize multiple times, i.e. does the `if is_drift` block need to be a `while is_drift` block?

src/factorizations/blocklanczos.jl
Jutho: I never know when to use `=` and `.=` when assigning a whole range of values, i.e.

```julia
state.H[(k + 1):(k + bs_next), (k - bs + 1):k] = view(B, 1:bs_next, 1:bs)
```

vs

```julia
state.R.vec[1:bs_next] .= Rnext.vec
```


 I think for these cases it doesn't matter, as long as you don't use `getindex` (which creates a copy) but only `view` in the right hand side.

src/factorizations/blocklanczos.jl
```diff
@@ -295,14 +301,17 @@ This function performs a QR factorization of a block of abstract vectors using t
 It takes as input a block of abstract vectors and a tolerance parameter, which is used to determine whether a vector is considered numerically zero.
 The operation is performed in-place, transforming the input block into a block of orthonormal vectors.
 
-The function returns a matrix of size `(r, p)` and a vector of indices goodidx. Here, `p` denotes the number of input vectors,
+The function returns a matrix of size `(r, p)`, a vector of indices goodidx and a boolean flag is_drift. Here, `p` denotes the number of input vectors,
```
Jutho:
```suggestion
The function returns a matrix of size `(r, p)`, a vector of indices `goodidx` and a boolean flag `is_drift`. Here, `p` denotes the number of input vectors,
```

src/factorizations/blocklanczos.jl
```diff
@@ -312,28 +321,43 @@ function block_qr!(block::Block, tol::Real)
...
+                if β < tol
+                    block[j] = zerovector!!(block[j])
+                    idx[j] = false
+                else
+                    R[j, j] = β
+                    block[j] = scale!!(block[j], 1 / β)
+                end
+            else
+                block[j] = scale!!(block[j], 1 / β)
+            end
```
Jutho: I think this whole block can be simplified to:

```suggestion
        if tol < β < 100 * tol # DGKS reorthogonalization
            is_drift = true
            for i in 1:(j - 1)
                δ = inner(block[i], block[j])
                R[i, j] += δ
                block[j] = add!!(block[j], block[i], -δ)
            end
            β = norm(block[j])
        end

        if β < tol
            block[j] = zerovector!!(block[j])
            idx[j] = false
        else
            R[j, j] = β
            block[j] = scale!!(block[j], 1 / β)
```

src/factorizations/blocklanczos.jl
Jutho:
```suggestion
    state.R.vec[1:bs_next] = Rnext.vec
```

src/factorizations/blocklanczos.jl
yuiyuiui: That logic is indeed cleaner. I have manually incorporated it into a new commit (the original suggestion had become stale and could not be applied directly due to recent changes).

src/factorizations/blocklanczos.jl
yuiyuiui: When `is_draft` is true, after a `block_reorthogonalize!`, `is_draft` can not be true again. So we do not need more `is_draft` check.

For the reason, although I can't exactly see when subspace draft happen, we can at least make sure that it happens when vectors in the block are near linear dependent. `block_qr!` guarantee that in its result, non-zero vector are exact linear independent (though it can not keep the origin subsapce). 

Then what `block_reorthogonalize!` does is to cut off vectors in $\text{span}(V)$, the error it makes in vector in supplementary space of $\text{span}(V)$ is just machine error and will not cause "near linear dependent". So we do not need a more `is_draft` check.

Jutho:
Thanks for this and especially for the long patience. I think this looks very good now, so I'll merge.

PR #151, opened 2026-02-06
[RFC] Symplectic Arnoldi

### LINKED ISSUE #150: support for skew-orthogonalization / symplectic Gram-Schmidt
I have a use case for an Arnoldi iteration, but where instead of being orthonormal, my `U` in `A * U = U * H` is actually symplectic, such that `Uᵀ * Jₘ * U = Jₙ` where `J₂ₙ = Iₙ ⊗ [0 1; -1 0]`. This can be achieved with a skew-symmetric version of the Gram-Schmidt process, such as described on pages 3-4 of https://people.math.ethz.ch/%7Eacannas/Papers/lsg.pdf. Most of the literature I could find on this only discusses symplectic Arnoldi methods in the context of Hamiltonian systems, but I am interested in calculating skew-orthogonal polynomials w.r.t. a skew-symmetric inner product, so my operator `A` is not Hamiltonian.

My question is, would something like this be in scope for KrylovKit.jl? Looking at the design in https://github.com/Jutho/KrylovKit.jl/blob/master/src/orthonormal.jl, I think it might lend itself quite nicely to extending it to different notions of "orthogonality" and there could then be different `SkewOrthogonalizer`s, which do different versions of symplectic Gram-Schmidt. Any thoughts regarding such a design? If that sounds interesting, I can try putting up a proof of concept.

simeonschaub:
See the discussion in #150. On my test case, this already gives me
significantly better performance, and improved accuracy, if using
re-skew-orthogonalization, than my hand-rolled Arnoldi implementation. An initial sketch was written by Copilot, but
I revised and reviewed the result extensively.

closes #150


simeonschaub:
Ok, I think I am happy with the implementation now from my side. This should enable some initial experimentation with symplectic Gram-Schmidt methods. Future work might include blocked methods such as described in https://journal.austms.org.au/ojs/index.php/ANZIAMJ/article/view/9380/1920, but I don't want to make this PR much bigger than it already is.

Jutho:
Ok, I will have to study this a bit further and let it sink in, but my initial impression is that I am not a big fan of "recycling" `inner` and `InnerProductVec` for computing the symplectic form, which than results in hacky solutions such as having to introduce a separate `norm` and a `standard_dot` function. While I agree that could duplication is typically harder to maintain, in this case I wouldn't mind if the symplectic functionality is provided through a separate `SymplecticFormVec` that copies a lot of the functionality of `InnerProductVec`, and uses a new dedicated `sympleticform` (or other shorter name) primitive for applying the symplectic form. I prefer the mathematical correctness of such an approach (`inner` and `symplecticform` have fundamentally different behavior under interchanging the two arguments), and think that will lead to cleaner and easier to comprehend code.

simeonschaub:
Sure, I can change the design! I wasn't too sure about extending `InnerProductVec` anyways, probably makes sense to introduce a new type for symplectic forms

src/factorizations/arnoldi.jl
```diff
@@ -47,13 +47,13 @@ Base.eltype(::Type{<:ArnoldiFactorization{<:Any, S}}) where {S} = S
 basis(F::ArnoldiFactorization) = F.V
 rayleighquotient(F::ArnoldiFactorization) = PackedHessenberg(F.H, F.k)
 residual(F::ArnoldiFactorization) = F.r
-@inbounds normres(F::ArnoldiFactorization) = abs(F.H[end])
+@inbounds normres(F::ArnoldiFactorization) = F.H[end]
```
Jutho: What is the rationale behind this change? Can `normres` be negative in the skew-orthogonal case? In the orthogonal case, I agree that `F.H[end]` is positive by construction, but in the complex case, it is a positive number encoded in a complex data type, and I would like `normres` to return a real number. So is `real(F.H[end])` a solution?

Jutho:
Thanks for the changes; this is starting to look really nice. I have a few remaining comments and suggestions:

1) I think it would be easier if `SymplecticFormVec` just lives in its own source file, similar to `InnerProductVec`.

2) Similarly, I would move everything related to `SymplecticBasis` and skew-orthogonalization in a separate file "skeworthonormal.jl", that lives next to "orthonormal.jl". 

3) I don't really like the `innerfun` keyword to `project`, since it is really completely determined by the type of `b`. I think this can more cleanly be solved by renaming `project` to `orthproject` and similarly have a `skewproject`, or simply by keeping the name `project` but have a separate method definition for `OrthonormalBasis` and `SymplecticBasis`. Again, I wouldn't mind that there is some code duplication here.

4) It would be great if you could at a bit more documentation, explaining that in order to use a `SkewOrthogonalizer` algorithm in `ArnoldiFactorization`, your vector type either needs to define `KrylovKit.symplecticform`, or declare which method is representing the symplectic form via `SimplecticFormVec(v, form)`, and also explaining the default implementation for `AbstractVector` types.

src/factorizations/arnoldi.jl
simeonschaub: Yes, for symplectic bases `normres` is allowed to be negative, since all the even vectors need to be scaled in a way such that `ω(last(b), v) = +1`. So if we don't allow `normres` to be negative `ω(last(b), v)` might come out to be -1 instead of +1, which is not what we want.

I haven't really played around with complex vectors yet, but I imagine the phase would be relevant there as well. Would a good compromise be to have two methods:

```julia
@inbounds normres(F::ArnoldiFactorization{<:Any, <:Any, <:OrthonormalBasis}) = abs(F.H[end])
@inbounds normres(F::ArnoldiFactorization{<:Any, <:Any, <:SymplecticBasis}) = F.H[end]
```

?

simeonschaub:
> It would be great if you could at a bit more documentation, explaining that in order to use a SkewOrthogonalizer algorithm in ArnoldiFactorization, your vector type either needs to define KrylovKit.symplecticform, or declare which method is representing the symplectic form via SimplecticFormVec(v, form), and also explaining the default implementation for AbstractVector types.

Did you want me to add some more docs regarding this to the docstrings of `ArnoldiFactorization` and `ArnoldiIteration` as well? I just thought since `InnerProductVec` isn't mentioned there either, it might be more appropriate to add that explanation to the docstrings of `SymplecticBasis` and the `(re)skeortho(gon|norm)alize(!!)` functions instead and link to that in the implementation details docs. I can add it there as well though if you think it makes sense.

simeonschaub:
Did you have a chance to take another look?

Jutho:
My apologies for the long silence. Overall I think this is a great PR with very nice and consistent coding patterns.

The only part of which I am not convinced is including both the orthogonal and symplectic Arnoldi process in the same `ArnoldiFactorization` object (by introducing the type of basis as a new type parameter), i.e. the changes in the "factorizations/arnoldi.jl" file, for two reasons. Conceptually, I think most people think of Arnoldi as having an orthogonal basis for the Krylov subspace (so the symplectic part is quite hidden in this type parameter). Codewise, I don't like the resulting `if ... else ... ` structures.

I would propose to have an explicit `SymplecticArnoldiFactorization` structure, and for those methods that would be identical for both structures, to use a `Union{ArnoldiFactorization,SymplecticArnoldiFactorization}` type specification for the factorization argument.

simeonschaub:
Sorry, I was a little busy with other stuff these last few months, but finally had time to come back to this! I addressed your review comments. I also had claude review the existing code and it found some correctness issues I had it fix and add some additional tests. I reviewed the fixes and they all seemed sensible to me

Issue #101, opened 2024-11-11
GPU acceleration via ITensors


YotamKa:
Hi,
I am simulating a quantum dynamical system using the great ITensorMPS.jl package.
(https://github.com/ITensor/ITensorMPS.jl)
Without getting into details about this package and the specific computation, I want to point out a seemingly significant overhead in computation when running a subroutine of the 1TDVP algorithm using "exponentiate" with a GPU backend. 
(it is used exactly here- https://github.com/ITensor/ITensorMPS.jl/blob/main/src/solvers/tdvp.jl)

This algorithm solves the (time-dependent-)Schrodinger equation by projecting the Hamiltonian onto a single MPS site. Then, the "exponentiate" function is used to solve the single-site effective equation. In principle, any ODE solver can solve the effective equation, but they usually work with exponentiate.

Is the significant overhead with the GPU backend compared with the CPU backend expected or known?

lkdvos:
Hi!

It would be really interesting if you could maybe share a profiler run of what you are trying to achieve. In principle, KrylovKit should be compatible with GPU-based backends, with the limitation that they still perform CPU linear algebra on the smaller matrices that characterize the Krylov subspaces. Depending on the interplay of these packages, it might be that there are accidentally a large number of transfers between CPU and GPU memory that are hindering the performance, or that it is some other thing that might be going wrong, or that it really is just that there is quite a bit of overhead that cannot be avoided.
Without further information, it's a bit hard to give a better answer.

YotamKa:
Thanks for the answer; it made me check my statement more closely.

At the end of this post, you can find a code built on example no.1 from ITensorMPS.jl.
(https://github.com/ITensor/ITensorMPS.jl/blob/main/examples/solvers/01_tdvp.jl)
This code performs imaginary time evolution to a 1D chain of spins defined by a Heisenberg Hamiltonian. As a function of the maximal bond dimension, which depends on the number of spins (n, in code), the CPU or GPU backends can be favorable. In larger problems (larger bond dimensions), the GPU performs better, as expected.

I have not tried to run a profile yet, but I will get to that soon.

It can be interesting to hear an opinion from someone more experienced with these considerations: for a given problem size, say n=20 and maxbondim=100, does the GPU backend have the expected speedup over the CPU (in my laptop, it was faster by around 0.75 per time step- 3sec compared with 4sec). 


Here's the code:
```
[...]

PR #154, opened 2026-03-07
Optimize project!! for GPU arrays by batching inner products (#101)

XingyuZhang2018:
## Summary

This PR fixes performance issue #101 where `eigsolve` with `Vector{CuArray}` on GPU was 100-200x slower than CPU equivalent.

## Root Cause

`project!!` was calling `inner()` sequentially for each Krylov basis vector, resulting in k×nc GPU dot operations (each requiring device synchronization). For `Vector{CuArray}` with k=30 basis vectors and 4 components: 120 serialized GPU kernel launches → `eigsolve` takes ~4.7s instead of <100ms on CPU.

## Solution

Introduce fast GPU paths in `src/orthonormal.jl`:

- `_use_matrix_array_kernel()`: detect non-CPU arrays (CuArray, etc.)
- `_use_batched_component_kernel()`: detect `Vector{non-CPU array}`
- `project_matrix!`: hcat basis vectors → single GEMV on GPU
- `project_batched_components!`: per-component batching for `Vector{CuArray}`
- `unproject_matrix!` / `unproject_batched_components!`: symmetric GPU paths for `unproject!!`

CPU and multithreaded paths are completely unaffected (dispatch predicates return `false` for `Array{<:Number}`).

## Performance Impact

For XingyuZhang2018's example (N=1000, nc=4, k=30):

- `project!!` single call: **3.8x faster** (9.2ms → 2.4ms)
- `eigsolve` total: **13.5x faster** (4.7s → 0.348s)
- CPU/multithreaded paths: **no regression** (original logic preserved)

## Testing

- ✅ CPU `Array` paths verified (original behavior)
- ✅ Multithreaded paths verified (original behavior)
- ✅ GPU `CuArray` paths verified with RTX 5060
- ✅ XingyuZhang2018's original example passes with 13.5x speedup

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/orthonormal.jl
+++ b/src/orthonormal.jl
@@ -72,6 +72,29 @@ function _use_multithreaded_array_kernel(::Type{<:OrthonormalBasis{T}}) where {T
     return _use_multithreaded_array_kernel(T)
 end
 
+# helper function to determine if matrix-based BLAS level-2 operations should be used
+# this is beneficial for non-CPU AbstractArrays (e.g., GPU arrays) where sequential
+# BLAS level-1 operations incur high per-call kernel launch overhead
+_use_matrix_array_kernel(y) = _use_matrix_array_kernel(typeof(y))
+_use_matrix_array_kernel(::Type) = false
+function _use_matrix_array_kernel(::Type{T}) where {T <: AbstractArray{<:Number}}
+    return !(T <: Array)
+end
+function _use_matrix_array_kernel(::Type{<:OrthonormalBasis{T}}) where {T}
+    return _use_matrix_array_kernel(T)
+end
+
+# helper for composite vectors (e.g., Vector of non-CPU arrays such as Vector{CuArray})
+# where inner products can be batched component-wise using matrix operations
+_use_batched_component_kernel(y) = _use_batched_component_kernel(typeof(y))
+_use_batched_component_kernel(::Type) = false
+function _use_batched_component_kernel(::Type{<:AbstractVector{T}}) where {T <: AbstractArray{<:Number}}
+    return _use_matrix_array_kernel(T)
+end
+function _use_batched_component_kernel(::Type{<:OrthonormalBasis{<:AbstractVector{T}}}) where {T <: AbstractArray{<:Number}}
+    return _use_matrix_array_kernel(T)
+end
+
 """
     project!!(y::AbstractVector, b::OrthonormalBasis, x,
         [α::Number = 1, β::Number = 0, r = Base.OneTo(length(b))])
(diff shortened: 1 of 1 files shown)


lkdvos (commented):
This is a good attempt, but I fear that it might not be so easy, mostly because we cannot actually assume anything about the `inner` function, which might not be defined as `y' * x`.
So at the very least, this would have to be restricted to the cases where we know `inner` reduces to the normal inner product on vectors.

I'm definitely not opposed to considering things like the definition of `batched_inner`, which we could then overload for performance cases?

Considering the concatenated `OrthonormalBasis`, it almost feels like in that case it might make sense to simply "go all the way", and consider introducing a separate basis type that stores all vectors contiguously, which would then only be chosen for types that support this behavior?
I don't know what you think about that, but would be happy to hear your opinions!

XingyuZhang2018:
Thanks for the review! You're absolutely right — we can't assume `inner` reduces to `dot`, and the original dispatch was not explicit enough about that.

I've just pushed a refactor that addresses this by introducing `batched_inner!` as the public extension point ([cc5865f](https://github.com/XingyuZhang2018/KrylovKit.jl/commit/cc5865f)):

**Design**
- `batched_inner!(y, b, x, α, β, r)` is a documented, overloadable function that computes `y[j] = β*y[j] + α*inner(b[r[j]], x)`.
- The **default method** contains all the dispatch logic:
  - For non-CPU `AbstractArray{<:Number}` types (e.g. `CuArray`): BLAS matrix-vector path.  These types use `VectorInterface.inner = dot`, so the optimisation is valid.
  - For `AbstractVector{<:AbstractArray{<:Number}}` (e.g. `Vector{CuArray}`): component-wise batching.
  - Everything else: sequential `inner` calls.  This correctly handles `InnerProductVec` (which is **not** `<: AbstractArray{<:Number}`) and any other wrapper with a custom inner product.
- `project!!` is simplified to: multithreaded-CPU branch **or** `batched_inner!`.
- Users who define a non-standard `inner` for a GPU-like type can override `batched_inner!` for that type to get both correctness and batching.

**Dead code removed**
`unproject_matrix!` and `unproject_batched_components!` were defined in the previous commit but were never actually dispatched to from `unproject!!`. Since sequential `add!!` / `axpy!` calls are already pipeline-friendly on GPU drivers (no synchronisation required between calls), there is no kernel-launch bottleneck there — only in `project!!` where each `inner` requires a device-to-host result transfer. I've removed the dead functions to keep the diff clean.

---

Regarding the **contiguous basis type** idea: I agree it would be a cleaner long-term solution — the type itself would encode the assertion "my inner product is the standard one and I store vectors contiguously", making the batched GEMV an obvious natural operation rather than a runtime type check. Happy to follow up with a separate PR for that if the direction looks good to you.

Jutho:
I think the way to approach is, is by first having a `batched_inner!` in VectorInterface.jl. In general, I would like `batched_inner!` to even support two sets of vectors, i.e. so that `batched_inner!(G, X, Y, α, β)` computes `G[i,j] = β * G[i,j] + α * inner(X[i], Y[j])`. I think this can have its use cases, for example to compute a [Gram(ian) matrix)` (namely when `X == Y`). 

The main issue I see is that, a non-mutating version `G = batched_inner(X, Y, α=1)` still has to be able to determine what the correct type of `G` needs to be. However, it might be possible to solve this using the recursive structure that VectorInterface is anyway using for all of its implementations.

Once this is in place, it should probably be possible to also fix KrylovKit more consistently and generically. However, we might want to think the interface through carefully, in order to avoid having to change the interfaces muliple times.

PR #158, opened 2026-06-01
Update Project.toml

kshyatt:
Allow use of VectorInterface 0.6


lkdvos:
This should really have been caught by the dependabot, any chance you want to take a look at adding that too (and removing compathelper?)

kshyatt:
ya can do in a bit, although I'm not a maintainer here!

